PR #1, opened 2022-07-09
Fix attributes aptitude contribution to statLifespan.

Lamphobic:
statLifespan is currently calculated from attribute value on reincarnation rather than what looks to be the intended attribute aptitude on reincarnation.


immortalityidle (approved):
Good catch. Thanks.

PR #9, opened 2022-07-13
Home improvements + Earlier suggestions

Hornsoup:
Current code overbuys land and only purchases one land per day. This should clean things up.

### DIFF
--- a/src/app/game-state/home.service.ts
+++ b/src/app/game-state/home.service.ts
@@ -474,7 +474,8 @@ export class HomeService {
         this.ageFields();
         if (this.home.costPerDay > this.characterService.characterState.money){
           this.logService.addLogMessage("You can't afford the upkeep on your home. Some thugs rough you up over the debt. You better get some money, fast.", "INJURY", 'EVENT');
-          this.characterService.characterState.status.health.value -= 20;
+          //Thugs want money and they can't collect if you're dead, so scare the player. But the rats and monsters want you dead, and this makes for easier pickings and it scales with health.
+          this.characterService.characterState.status.health.value *= 0.5;
           this.characterService.characterState.money = 0;
         } else {
           this.characterService.characterState.money -= this.home.costPerDay;
@@ -557,7 +558,7 @@ export class HomeService {
   }
 
   upgradeToNextHome(){
-    if (this.houseBuildingProgress < 1){
+    if (this.upgrading){
       // currently upgrading, bail out
       return;
     }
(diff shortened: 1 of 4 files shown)


immortalityidle (commented):
I appreciate the effort here. Could you please separate out the home autobuy stuff from the rest? Smaller and more focused pull requests make it a lot easier for me to test and make sure it's what I want in the game. The middle tick solution isn't the way I want to solve the too-slow merging problem, and the bloodline money multiplier is intentional as a way to make money obsolete at that point in the game.

Hornsoup:
Still figuring out github since I haven't used a repository in like a decade, should be just the home.service changes branch now after reverting the bloodline and faster tick.

src/app/game-state/home.service.ts
```diff
@@ -474,7 +474,8 @@ export class HomeService {
         this.ageFields();
         if (this.home.costPerDay > this.characterService.characterState.money){
           this.logService.addLogMessage("You can't afford the upkeep on your home. Some thugs rough you up over the debt. You better get some money, fast.", "INJURY", 'EVENT');
-          this.characterService.characterState.status.health.value -= 20;
+          //Thugs want money and they can't collect if you're dead, so scare the player. But the rats and monsters want you dead, and this makes for easier pickings and it scales with health.
+          this.characterService.characterState.status.health.value *= 0.5;
```
immortalityidle, line 478: Let's leave this change out. In the early game I want the feel of constant danger, and this would remove a lot of that. It would also be very different from what I want in the late game. When you have a lot of health and health recovery, you can mostly shrug off the thugs, allowing strategies like growing your spirituality through fasting.

src/app/game-state/home.service.ts
```diff
@@ -698,35 +699,52 @@ export class HomeService {
   }
 
   autoBuy(){
+    let priceBuffer = (this.home.costPerDay + 1) * 14; // Two weeks, by popular request.
```
immortalityidle, line 702: Let's compromise on ten days.

src/app/game-state/home.service.ts
```diff
@@ -698,35 +699,52 @@ export class HomeService {
...
     if (this.autoBuyHomeUnlocked && this.homeValue < this.autoBuyHomeLimit){
+      // check for final building to avoid buying excess building land.
+      if (!this.upgrading && this.nexthome.HomeType < this.autoBuyHomeLimit){
```
immortalityidle, line 705: Why the extra check on the home limit? Does the if clause above not already do that?

src/app/game-state/home.service.ts
```diff
@@ -698,35 +699,52 @@ export class HomeService {
...
-        }
       }
     }
-    if (this.autoBuyLandUnlocked && (this.land + this.fields.length + this.extraFields) < this.autoBuyLandLimit){
-      //autoBuy is on, check if we have the money for the land plus the next few days' rent and food
-      if (this.characterService.characterState.money > this.landPrice  + (this.nextHome.costPerDay * 3) + 3 ){
+    // if there's no autohome, autohome is finished, or there's enough money to cover buying a new plot of land and the home, try buying land.
+    if (!this.autoBuyHomeUnlocked ||
+        this.homeValue >= this.autoBuyHomeLimit ||
+        (this.characterService.characterState.money >= this.nextHome.cost + this.landPrice + priceBuffer )){
```
immortalityidle, line 724: Seems like we should just check the money in the while loop condition.

src/app/game-state/home.service.ts
```diff
@@ -698,35 +699,52 @@ export class HomeService {
...
+      //keep checking if we have the money for the land plus food and rent.
+      while (this.autoBuyLandUnlocked &&
```
immortalityidle, line 726: We should just check the unlocks in the if above.

src/app/game-state/home.service.ts
Hornsoup: Because this is for the next home, not the current home, and I wasn't sure if you could just +1 the value. If I got it right, the logic *should* be that you only buy land if you need it to build the next home, or that you're not already building the final home. Upon review I think the not needs to be outside a parenthetical for this to make sense, there's so much logic going on in here I can't be sure.

Current behavior is that while you're upgrading the final home in the buy limit, you buy land for it even though you don't need it. A tiny issue unless you decide to change the price scaling of land in the future. Honestly I'd suggest doing away with buying land and make it purely fields and a price for home since empty lots aren't used for anything I've seen, unless I'm missing something in impossible tasks.

Hornsoup: And now I realize that check isn't needed because if you're not upgrading you can't get past that first check if you're done. I blame the sleep deprivation.

src/app/game-state/home.service.ts
Hornsoup: It's an || statement, so without this check it won't buy anything if you have autohome and you haven't reached your buy limit. This makes sure you have more than enough for the next home and allows you to autobuy farmland during a build.

src/app/game-state/home.service.ts
Hornsoup: Removed.

src/app/game-state/home.service.ts
```diff
@@ -699,11 +698,11 @@ export class HomeService {
   }
 
   autoBuy(){
-    let priceBuffer = (this.home.costPerDay + 1) * 14; // Two weeks, by popular request.
+    let priceBuffer = (this.home.costPerDay + 1) * 10; // Ten days, by popular request.
```
Hornsoup, line 701: Changed to 10 days.

src/app/game-state/home.service.ts
```diff
@@ -474,8 +474,7 @@ export class HomeService {
         this.ageFields();
         if (this.home.costPerDay > this.characterService.characterState.money){
           this.logService.addLogMessage("You can't afford the upkeep on your home. Some thugs rough you up over the debt. You better get some money, fast.", "INJURY", 'EVENT');
-          //Thugs want money and they can't collect if you're dead, so scare the player. But the rats and monsters want you dead, and this makes for easier pickings and it scales with health.
-          this.characterService.characterState.status.health.value *= 0.5;
+          this.characterService.characterState.status.health.value -= 20;
```
Hornsoup, line 477: Reverted.

src/app/game-state/home.service.ts
```diff
@@ -699,11 +698,11 @@ export class HomeService {
...
 
   autoBuy(){
-    let priceBuffer = (this.home.costPerDay + 1) * 14; // Two weeks, by popular request.
+    let priceBuffer = (this.home.costPerDay + 1) * 10; // Ten days, by popular request.
     if (this.autoBuyHomeUnlocked && this.homeValue < this.autoBuyHomeLimit){
-      // check for final building to avoid buying excess building land.
-      if (!this.upgrading && this.nexthome.HomeType < this.autoBuyHomeLimit){
-        //try to buy land, and do it instantly.
+      // Don't buy land while upgrading.
+      if (!this.upgrading){
```
Hornsoup, line 704: Removed the check, it really wasn't needed...

src/app/game-state/home.service.ts
```diff
@@ -719,13 +718,13 @@ export class HomeService {
       }
     }
     // if there's no autohome, autohome is finished, or there's enough money to cover buying a new plot of land and the home, try buying land.
-    if (!this.autoBuyHomeUnlocked ||
+    if ((!this.autoBuyHomeUnlocked ||
         this.homeValue >= this.autoBuyHomeLimit ||
-        (this.characterService.characterState.money >= this.nextHome.cost + this.landPrice + priceBuffer )){
+        (this.characterService.characterState.money >= this.nextHome.cost + this.landPrice + priceBuffer )) &&
+        this.autoBuyLandUnlocked){
```
Hornsoup, line 724: Moved the unlock

immortalityidle (approved):

immortalityidle:
Resolve the merge conflicts and I think this is good to go.

Hornsoup:
Github hurt intself in its confusion.

PR #10, opened 2022-07-13
Builders help tower

Hornsoup:
It's messy with merges and reverts, but it's just adding a for loop of builder x builder.power / 10 to tower parts creation to reduce the 547,945 years of brick building.


immortalityidle (commented):
I don't think this is the way I want to address the complaint that the tower takes too long. With this change the player could level up a bunch of builders and make the whole tower in a very short time. It wouldn't feel epic enough. I was thinking instead that we could solve the problem by adding a miner worker that collects ore.

src/app/game-state/activity.service.ts
```diff
@@ -542,11 +542,22 @@ export class ActivityService {
       consequence: [() => {
         this.characterService.characterState.status.stamina.value -= 100;
         let oreValue = 0;
+        let builderPower = 0;
         for (let i = 0; i < 200; i++){
           oreValue = this.inventoryService.consume('ore');
         }
+        for (let follower of this.followerService.followers){
+          if (follower.job = "builder"){
```
immortalityidle, line 550: You need to be careful  with == vs =. This loop would turn all your followers into builders every time it ran.

src/app/game-state/character.ts
```diff
@@ -282,6 +282,9 @@ export class Character {
     if (this.money > this.maxMoney){
       this.money = this.maxMoney;
     }
+    if (this.money > this.maxMoney){
```
immortalityidle, line 285: Not sure why this code chunk is duplicated here.

Hornsoup: No idea either, probably a bad merge somewhere.

src/app/game-state/activity.service.ts
Hornsoup: Ooh, that's bad, because I think I copy pasted that from something, I'll doublecheck where I got it.

Hornsoup: Yeah, check line 631 of the base: https://github.com/Hornsoup/immortalityidle.github.io/blob/26b2602eea164931b5cffbdcd54737ff1df789f8/src/app/game-state/activity.service.ts#L619

Hornsoup:
> I was thinking instead that we could solve the problem by adding a miner worker that collects ore.
This would actually make it even faster than my implementation, because if a miner collects 1 ore per day per power they'd be going at the rate 10 builders help with the tower and you'd feel more like a manager than a team.
It would also 
- add a lot more inventory add cycles
- create another follower type to reject unless you're making weapons
- and you'd have to clear them out and swap in builders in order to make the bricks. 
If you want this for your design it's your prerogative, I just want to put forth all the issues.

src/app/game-state/activity.service.ts
immortalityidle: Hahahaha! What a fun way to catch a bug.

immortalityidle:
Good points. Let me think about it for a bit.

immortalityidle:
OK. Thought about it. How about we remove the scaffolding help and decrease the effect of follower power on bricks and mortar to 1/10th?

src/app/game-state/activity.service.ts
```diff
@@ -579,11 +579,23 @@ export class ActivityService {
...
+            builderPower += follower.power;
+          }
+        }
+        builderPower = Math.floor(builderPower /10);
         if (oreValue >= 10){
           this.inventoryService.addItem(this.itemRepoService.items['everlastingBrick']);
+          for (let i = 0; i < builderPower; i++){
+            this.inventoryService.addItem(this.itemRepoService.items['everlastingBrick']);
+          }
+          this.logService.addLogMessage("You and your followers made " + (1 + builderPower) + " " + this.itemRepoService.items['everlastingBrick'].name,"STANDARD","CRAFTING");;
```
immortalityidle, line 597: Are we off by one here? Looks like we create builderPower bricks.

src/app/game-state/activity.service.ts
```diff
@@ -579,11 +579,23 @@ export class ActivityService {
       consequence: [() => {
         this.characterService.characterState.status.stamina.value -= 100;
         let oreValue = 0;
+        let builderPower = 0;
```
immortalityidle, line 582: Start this at 1 for the one brick we can make ourselves?

src/app/game-state/activity.service.ts
Hornsoup: Oop, yeah that's an oversight.

Hornsoup:
Technically didn't need to since there was an initial inventory add that I didn't surround, but did it anyway to avoid overlooking.

PR #11, opened 2022-07-14
Balancing Death Fight

Hornsoup:
Updated combat mechanics so that armor and defense don't make things become a joke, added in followers adding weapon damage and armor so you have a chance to kill Death within this lifetime.


src/app/game-state/followers.service.ts
```diff
@@ -75,10 +75,14 @@ export class FollowersService {
         if (this.characterService.characterState.equipment.rightHand && 
           this.characterService.characterState.equipment.rightHand.weaponStats){
           this.characterService.characterState.equipment.rightHand.weaponStats.durability += follower.power;
+          this.characterService.characterState.equipment.rightHand.weaponStats.baseDamage += Math.floor(follower.power/10);
```
Hornsoup, line 78: This new damage (and value) added by followers should make it possible to improve weapons fast enough to win against Death.

src/app/game-state/battle.service.ts
```diff
@@ -284,7 +287,7 @@ export class BattleService {
       maxHealth: this.troubleKills * 10,
       accuracy: 0.5,
       attack: this.troubleKills / 5,
-      defense: Math.floor(Math.log2(this.troubleKills)),
+      defense: this.troubleKills / 5,
```
Hornsoup, line 290: Defense for trouble needs to keep up with the new combat equation.

src/app/game-state/battle.service.ts
```diff
@@ -144,7 +144,11 @@ export class BattleService {
       for (let i = 0; i < enemyStack.quantity; i++){
         if (Math.random() < enemyStack.enemy.accuracy){
           let damage = enemyStack.enemy.attack;
-          damage = damage / (1 + this.characterService.characterState.defense * 0.1);
+          let combatDefense = this.characterService.characterState.defense;
```
Hornsoup, line 147: The old equation would make defense make Death's effective health in the 18th order of magnitude, far beyond what players can make weapons do in a real-time century.

src/app/game-state/battle.service.ts
```diff
@@ -144,7 +144,11 @@ export class BattleService {
...
+          // with 20k max health currently, the curve has to be related to Death's damage so player has the chance to survive and oneshot Death.
+          damage = damage / (1 + Math.pow(100 + this.enemyRepo.death.attack / 20000, (-damage + combatDefense) / combatDefense) );
+          //Keep mice scary
+          if (damage < this.enemyRepo.mouse.attack) damage = this.enemyRepo.mouse.attack;
```
Hornsoup, line 151: The old equation would make defense make Death's effective health in the 18th order of magnitude, far beyond what players can make weapons do in a real-time century.

src/app/game-state/followers.service.ts
```diff
@@ -75,10 +75,14 @@ export class FollowersService {
...
+          this.characterService.characterState.equipment.rightHand.value += Math.floor(follower.power/10);
```
Hornsoup, line 79: This new damage (and value) added by followers should make it possible to improve weapons fast enough to win against Death.

immortalityidle:
What kind of timeframe does this put the death kill at, assuming the player is past softcaps on everything?

Hornsoup:
Since the player needs enough damage to oneshot Death in this scenario they'd be spending a large amount of time just improving weapons to meet the health. With a team of 20 Weaponsmiths you'd average 1 skill increase every 5 days, 50 days for a +1 to weapon per day from Weaponsmiths, outpacing the player in doing smithing after the team exists for 8 years assuming player is using an improved automerge, without which they'll probably not bother. This means every 8 years a team of 20 exists they'll be able to add damage equivalent to a new 130 ish damage weapon per day, but this is in line with the increase in the Durability. Possibly the power increase per member needs a lower chance at higher levels?

Death has a health of 100 million, and players take several hours per million currently if done by hand waiting for the current automerge. If you want death to be unkillable for now it'd be best to divide by a higher max health so it oneshots the player and add a max health increase when ready.

src/app/game-state/followers.service.ts
```diff
@@ -88,18 +92,26 @@ export class FollowersService {
         if (this.characterService.characterState.equipment.head && 
           this.characterService.characterState.equipment.head.armorStats){
           this.characterService.characterState.equipment.head.armorStats.durability += follower.power;
+          this.characterService.characterState.equipment.head.armorStats.defense += Math.floor(follower.power/10);
```
immortalityidle, line 95: This section ended up getting really bulky. Can we just loop through the slots so we have less duplicated code?

src/app/game-state/battle.service.ts
```diff
@@ -144,7 +144,11 @@ export class BattleService {
       for (let i = 0; i < enemyStack.quantity; i++){
         if (Math.random() < enemyStack.enemy.accuracy){
           let damage = enemyStack.enemy.attack;
-          damage = damage / (1 + this.characterService.characterState.defense * 0.1);
+          let combatDefense = this.characterService.characterState.defense;
+          // with 20k max health currently, the curve has to be related to Death's damage so player has the chance to survive and oneshot Death.
+          if (combatDefense > 0) damage = damage / (1 + Math.pow(100 + this.enemyRepo.death.attack / 20000, (-damage + combatDefense) / combatDefense) );
```
immortalityidle, line 149: Nitpick: bracket all if statements. I know one-liners are legal, but it's the way all the rest of the codebase is and I'd like to keep it consistent for better readability.

src/app/game-state/followers.service.ts
Hornsoup: I'll try doing that, have to wrap my head around the specific logic of how.

Hornsoup: Okay, so I want to do this, but I have no idea how to loop the slot since it's part of characterService and not inventoryService.

Hornsoup:
I *think* I understand how using Object.keys() works, I'm not 100%. I'm inferring a lot about how you're using polymorphism elsewhere.

immortalityidle:
With the recent improvements to automerging is the climb to get gear good enough for the death fight still comparable? Also, I'm not loving the idea of the final fight always needing to be a one-shot kill (in either direction). We can play with death's stats, but I'd really like it to be a little more drawn out and still require really high end gear.

Hornsoup:
From what I've seen from players, It still requires a lot of high end gear with this, though Death health might need increasing for an epic fight (The current fight is basically just a huge health bar with how defense works so 1mil x health is all you'd need), and I was thinking that an alternative is enemies / player could have some small amount of natural regen under 1%? 

The big factor with this pull is whether the player can heal 20k health between fight rounds, which seems to be possible currently with a huge amount of farms assuming they're spread out. If you'd rather remove the damage minimum so the player isn't constantly on the verge of death with high enough defense I can remove the Defense divider so the number can be put to a power higher than 1.

Hornsoup:
Made some changes so that players will need a small team and gear to get this done, scouts for fast attacks, hunters to provide meat to heal you, over 1 billion attack and defense (players are reaching about 100 million fairly quickly, it *should* take a few realtime days even with weapon and armor smith updates) and nearly 85k lifespan. The fight will last about 3 days in realtime under these conditions, possibly shorter with weaponsmiths continuously improving the weapon.

Hornsoup:
Tested the start of the fight, and it is reprehensibly long and requires at least 750 million defense from the player to not be killed. Got that in a few hours, the fight itself should take a day or two of watching numbers go down.

src/app/game-state/battle.service.ts
```diff
@@ -143,7 +143,15 @@ export class BattleService {
       for (let i = 0; i < enemyStack.quantity; i++){
         if (Math.random() < enemyStack.enemy.accuracy){
           let damage = enemyStack.enemy.attack;
-          damage = damage / (1 + this.characterService.characterState.defense * 0.1);
+          let defense = this.characterService.characterState.defense;
+          // with 20k max health currently, the curve has to be related to Death's damage so player has the chance to survive and oneshot Death.
```
immortalityidle, line 147: We're not planning to require a one-shot kill now, right? Want to update the comment?

src/app/game-state/battle.service.ts
```diff
@@ -325,27 +336,27 @@ export class BattleService {
...
       name: "an angry army",
-      health: 2000000,
-      maxHealth: 20000000,
+      health: 20000000,
+      maxHealth: 20000000, 
       accuracy: 0.9,
-      attack: 1000000,
-      defense: 500000,
+      attack: 10000000,
+      defense: 10000000, // Army was a bit weak when I accidentally fought them, increased defense should help.
```
immortalityidle, line 350: Appreciate the FYI but don't think the comment needs to stay in the code.

immortalityidle (commented):
A couple of nitpicks on the comments and them I think this is good to go.

src/app/game-state/battle.service.ts
Hornsoup: I mostly do that for the review so that when I'm coding at 2am you can see a comment from me when I forget to add comments in git from sleep deprivation and don't think about how this is permanently in code.

Hornsoup:
Should be good.

immortalityidle (approved):

PR #13, opened 2022-07-14
Just Keep Merging

Hornsoup:
Don't return, just get the item you just merged and look for the next. Automerge everything instantly.

### DIFF
--- a/src/app/game-state/inventory.service.ts
+++ b/src/app/game-state/inventory.service.ts
@@ -1060,7 +1060,9 @@ export class InventoryService {
                         destinationItem = item;
                     }
                   }
-                } else mergeDestinationIndex = -1;
+                } else {
+                  mergeDestinationIndex = -1;
+                }
               }
             }
           }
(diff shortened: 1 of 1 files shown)


src/app/game-state/inventory.service.ts
```diff
@@ -1053,7 +1053,14 @@ export class InventoryService {
                 this.itemStacks[mergeDestinationIndex] = null;
                 this.itemStacks[i] = null;
                 this.mergeEquipment(destinationItem, sourceItem, mergeDestinationIndex );
-                return;
+                item = this.itemStacks[mergeDestinationIndex]?.item;
+                if (item){
```
immortalityidle, line 1057: Do we need these checks? Seems like we could just skip to setting destinationItem=item.

Hornsoup: It's part of best practices since there's nothing to ensure that the destination hasn't been nulled during the mergeEquipment call.

PR #20, opened 2022-07-15
Health panel progress bars

Ch1pless:
Swapped progress element for a more generic but cross platform div and span combo.


immortalityidle:
I like this. We should probably do this all over where we use progress elements. Want to move the style stuff into the styles.less file so we don't have to replicate it in all the component less files?

Ch1pless:
Yes I can do so, I'll do that some point today. Do you think making the text display "value / max" would be something to do too? Or leave it as "value"?

Ch1pless:
Also, should I apply the Math.floor to all display values considering the recent bug on the discord?

Kleptichron:
I think having max would be preferable from a player standpoint. There's plenty of times where stamina or mana are just hovering at the bottom of the range and the max isn't visible anywhere else.

immortalityidle:
I think the nutrition bar might still have issues in dark mode, white text on yellow. If you flip to light mode does the dark text show up on the dark background? or does the background of the widget flip with the mode?

immortalityidle:
I'm also not married to the nutrition color. If it makes your life easier to change that, that's fine.

Ch1pless:
the modification to app.component.less was spacing formatted (some css classes had a 2 space indent, others had a 4 space one)

Ch1pless:
Just to make sure Credit to  @Coebh for the gradient backgrounded text on the progress bars.

immortalityidle (approved):
So excited to test this out.

PR #21, opened 2022-07-15
Removing loops from land purchase.

Hornsoup:
Added a method for calculating how much land a certain amount of money can buy (potentially rolling max land update on button!) and updated all the methods to use math instead of loops.


src/app/game-state/home.service.ts
```diff
@@ -717,23 +750,33 @@ export class HomeService {
...
     // if there's no autohome, autohome is finished, or there's enough money to cover buying a new plot of land and the home, try buying land.
     if ((!this.autoBuyHomeUnlocked ||
         this.homeValue >= this.autoBuyHomeLimit ||
         (this.characterService.characterState.money >= this.nextHome.cost + this.landPrice + priceBuffer )) &&
         this.autoBuyLandUnlocked){
-      //keep checking if we have the money for the land plus food and rent.
-      while ((this.land + this.fields.length + this.extraFields) < this.autoBuyLandLimit &&
-             (this.characterService.characterState.money >= this.landPrice + priceBuffer) ){
-        //break if reduced to money for home.
+      //check if we have the money for the land plus food and rent.
```
immortalityidle, line 759: Didn't we already do this code?

src/app/game-state/home.service.ts
```diff
@@ -689,23 +689,56 @@ export class HomeService {
...
+
+  /**
+   * 
+   * @param money the money available for use
+   * @returns count of affordable land
+   */
+  calculateAffordableLand(money: number): number{
+    let x = money;
+    let C = this.landPrice;
+    return Math.floor(((-C - 5) + (Math.sqrt(Math.pow(C, 2) + 10 * C + 20 * x + 25)))/10); // I know this looks nuts but I tested it on its own ^_^;;
```
immortalityidle, line 726: Shouldn't the affordable amount include a buffer so they don't go fully broke?

src/app/game-state/home.service.ts
Hornsoup: This is part of removing the loop, though I'm not 100% on the logic here because it gets hectic, it should be good.

Hornsoup: Because we still want to check to see if they're building or saving for a home before we calculate the money that's available to use for automatic land purchase.

immortalityidle: You sound a little hesitant. Have you tested this?

Hornsoup: Tested it. Works exactly as I expect.

immortalityidle (approved):

PR #22, opened 2022-07-15
Auto-buy reserve options

Kleptichron:
image
image

Adds a checkbox to enable the static auto-buy reserve, and an input element to input the amount to be reserved. If the box is unchecked, the reserve amount is ignored and 10 * daily expenses is used. If checked, it uses the input instead of the base reserve amount.

### DIFF
--- a/src/app/options-modal/options-modal.component.html
+++ b/src/app/options-modal/options-modal.component.html
@@ -22,6 +22,14 @@
       </select>
     </span>
   </div>
+  <div *ngIf="homeService.autoBuyLandUnlocked || homeService.autoBuyHomeUnlocked || homeService.autoBuyFurnitureUnlocked" class="optionField">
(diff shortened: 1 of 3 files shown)


src/app/options-modal/options-modal.component.html
```diff
@@ -22,6 +22,14 @@
       </select>
     </span>
   </div>
+  <div *ngIf="homeService.autoBuyLandUnlocked || homeService.autoBuyHomeUnlocked || homeService.autoBuyFurnitureUnlocked" class="optionField">
```
Kleptichron, line 25: Appears below auto-buy options if any auto-buy manual has been purchased. For furniture, this looks a little weird since it doesn't actually add any options.

immortalityidle (approved):

PR #24, opened 2022-07-15
Progressively auto-buy furniture option

Kleptichron:
image

https://user-images.githubusercontent.com/109362801/179318899-efb8b504-60e4-4ccb-a93e-52358849e12c.mov

Adds a new option to buy furniture progressively rather than waiting until we can afford the last one we bought manually.
It will purchase successively until we reach the last manually bought furniture piece, even if that's the worst one.

This introduces the property "previousId" to furniture items to allow for a way to find the previous "tier" item.

With this option off, I _think_ everything will function identically to how it did before, but there are a few cases to test around auto-buying priority, like if this will buy the chef's kitchen instead of a Mansion during an upgrade to Manor.


src/app/game-state/home.service.ts
```diff
@@ -720,9 +724,6 @@ export class HomeService {
         // autoBuy is on, we have enough land, check if we have the money for the house plus food and rent
         if ((this.characterService.characterState.money > this.nextHome.cost + priceBuffer )){
           this.upgradeToNextHome();
-        } else {
```
Kleptichron, line 723: It _seems_ like this isn't necessary, since the other things check for it first, and because house is first in the list here. Returning here would prevent buying, say, a 10 tael blanket when we don't have enough money to purchase a mansion.

src/app/game-state/home.service.ts
```diff
@@ -758,15 +759,31 @@ export class HomeService {
...
+                this.characterService.characterState.money -= thingToBuy.value;
+                this.ownedFurniture.push(thingToBuy.name);
+                this.furniture[slot] = this.itemRepoService.getFurnitureById(thingToBuy.id);
+                break;
+              } else {
+                thingToBuy = this.itemRepoService.getFurnitureById(thingToBuy.previousId ?? '');
+              }
+            }
+          }
+        } else {
```
Kleptichron, line 776: Same logic as before from here on.

src/app/game-state/inventory.service.ts
```diff
@@ -50,7 +50,8 @@ export interface Pill extends Item {
 }
 
 export interface Furniture extends Item {
-  slot: FurniturePosition
+  slot: FurniturePosition,
+  previousId?: string
```
Kleptichron, line 54: Using previous ID instead of some static tier takes care of weird cases like bed of nails. It isn't previous to heated bed, so it will be skipped when progressively upgrading.

Still, any suggestions on a better way to store and handle this are welcome!

Kleptichron: Also, it being optional allows for items that don't have previous "iterations", like all workshop items.

src/app/game-state/item-repo.service.ts
```diff
@@ -52,11 +53,12 @@ export class ItemRepoService {
       type: 'furniture',
       slot: 'bed',
       value: 10000,
+      previousId: 'mat',
       description: "A fine bed with a cover. Curtains keep the mosquitoes off you during the night. Increases daily stamina recovery by 2 and restores a bit of health.",
       useConsumes: false,
       use: () => {
         this.characterService.characterState.status.stamina.value += 2;
-        this.characterService.characterState.status.health.value += 0.02;
+        this.characterService.characterState.status.health.value += 0.2;
```
Kleptichron, line 61: This _seems_ like a typo since the previous bed (blanket) restores 0.1. Let me know @immortalityidle

immortalityidle: Yeah, that was just me being sloppy.

src/app/game-state/inventory.service.ts
immortalityidle: Seems like it should work. I can't think of a better way off the top of my head.

src/app/game-state/home.service.ts
immortalityidle: Yeah, I think I wrote this part before furniture existed and never updated it. Kind of feeds into the feature request to allow the user to specify the autobuy priorities. I'm fine with removing the return for now.

ChandlerFerry:
@Kleptichron For my knowledge, do you plan on finishing this work?

@immortalityidle I should also ask, is this feature something that you want implemented?

Kleptichron:
Yes - working on a more comprehensive auto-buyer change that allows for priority choice.

Issue #23, opened 2022-07-15
"Pause on Death" doesn't immediately stop.


OneOfMany07:
As the title says, when using one of the higher speeds the pause on death feature doesn't trigger until after some work was done.

I've seen this a number of times using the highest speed I've unlocked (3rd box after pause... play with line on right).  On the next life I've had one turn happen.  I swear I even saw this but no days (today I was at 18 yrs 1 day when it paused).  I've seen it lock away other apprentice choices.  This time I had a Gathering Herbs trigger (2 rotten patchouli in inventory with the other after reincarnation items I expect).

Happens both on Chrome and Firefox.  v1.0.36 game version displayed

immortalityidle:
Should be fixed in the latest build.

PR #27, opened 2022-07-16
#23 Pause on death should stop without starting the next life

illo210:
Stop death consumning ticks of the next life by:
 - During the iteration of the ticks in main loop service if the game is paused during a tick, the nexts are not executed and saved in the banked ticks instead.
 - During the reincarnation in character service moving the actual reincarnation of the player to the next tick for stopping the next observables called during this tick of executing.

### DIFF
--- a/src/app/game-state/character.service.ts
+++ b/src/app/game-state/character.service.ts
@@ -5,6 +5,7 @@ import { ReincarnationService } from './reincarnation.service';
 import { Character, AttributeType } from './character';
 import { formatNumber, TitleCasePipe } from '@angular/common';
 import { ActivityService } from './activity.service';
+import { Subscription } from "rxjs";
 
 @Injectable({
   providedIn: 'root'
@@ -15,6 +16,7 @@ export class CharacterService {
   forceRebirth: boolean = false;
   fatherGift: boolean = false;
   lifespanTooltip: string = "";
+  deathSubscriber?: Subscription;
 
   constructor(
     private injector: Injector,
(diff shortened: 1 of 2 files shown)


src/app/game-state/main-loop.service.ts
```diff
@@ -100,7 +100,11 @@ export class MainLoopService {
           this.tickCount++;
           if (this.tickCount >= this.tickDivider){
             this.tickCount = 0;
-            this.tick();
+            if (this.pause) {
```
immortalityidle, line 103: I'm medium sure that you can't get to this point in the code if the pause flag is set. Did you test this?

immortalityidle: Oh, wait. I could be very wrong since the subscribers can set the pause flag.

immortalityidle:
@jimboradley want to take a look at this since you did the initial subscription mojo?

src/app/game-state/main-loop.service.ts
ChandlerFerry: image

Based on my testing, I believe this does indeed function as intended.

jimboradley:
I'll have to try this out when I have a machine to run it. I guess from looking at the code I don't really get how this is supposed to work

src/app/game-state/main-loop.service.ts
Hornsoup: Anecdote: I've always had at least 3 days added after death. I'm stopping at 0 every time on this PR, so the fix definitely works.

src/app/game-state/character.service.ts
```diff
@@ -61,7 +63,11 @@ export class CharacterService {
         this.characterState.dead = true;
         this.characterState.reincarnate(); // make sure character reincarnation fires before other things reset
         this.reincarnationService.reincarnate();
-        this.characterState.dead = false;
+        // Revive the character in the next tick update for making sure that everything is stopped.
+        this.deathSubscriber = this.mainLoopService.tickSubject.subscribe(() => {
```
jimboradley, line 67: So, I see death subscriber getting subscribed to, but I don't ever see a next() getting called for it. Seems to me this subject does nothing. Granted I'm looking at this on my phone, but what am I missing?

immortalityidle (approved):
Will test this too later today and put it in the next version unless @jimboradley beats me to it.

PR #28, opened 2022-07-16
Shift Click when adding activities to start at 10

Croebh:
This allows the user to Shift Click on activities to add them with 10 days to start with, instead of just the base 1.

### DIFF
--- a/src/app/activity-panel/activity-panel.component.html
+++ b/src/app/activity-panel/activity-panel.component.html
@@ -10,7 +10,10 @@
       <tr *ngIf="(activity.unlocked)" draggable="true" (dragstart)="drag(activity, $event)">
         <td>
           <span class="noWrap">
-            <button (click)="onClick(activity)" tooltip="Add this activity to the bottom of your queue. Shift-click to start at 10 days.">{{activity.name[activity.level]}}</button>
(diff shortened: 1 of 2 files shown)


src/app/activity-panel/activity-panel.component.html
```diff
@@ -10,7 +10,10 @@ <h3 class="panelHeader">
       <tr *ngIf="(activity.unlocked)" draggable="true" (dragstart)="drag(activity, $event)">
         <td>
           <span class="noWrap">
-            <button (click)="onClick(activity)">{{activity.name[activity.level]}}</button>
+            <button (click)="onClick(activity, $event)" 
+                    tooltip="Add this activity to the bottom of your queue. Shift-click to start at 10 days.">
```
immortalityidle, line 14: Maybe "Shift click to also repeat it for 10 days."? And I think we're calling it a schedule rather than a queue most places where it's referenced, but double check that.

Croebh: Valid, lemme check and update this

Croebh: It says schedule in the button to remove items from the queue, so went ahead with that change

immortalityidle (approved):

PR #35, opened 2022-07-17
Fix farming

Hornsoup:


Hornsoup:
It's causing an infinite loop right now and a quick review of the code doesn't elucidate why.

Hornsoup:
Because I didn't actually apply it to the button I was using. Fixed.

ChandlerFerry (approved):

PR #37, opened 2022-07-17
fix: attribute scaling math above softcap

ChandlerFerry:
[Original bug from Cho Quan](https://discord.com/channels/996414713766363146/996488799938936922/998133259714183229)

Math double checked and I came up with a slightly different implementation that closer matches current implementation.

Concern is that many players are progressing far with spirituality scaling with the wrong aptitude multiplier, so it is treated as a "feature" not a "bug".

### DIFF
--- a/src/app/game-state/character.ts
+++ b/src/app/game-state/character.ts
@@ -374,7 +374,7 @@ export class Character {
       // increase by log2 of whatever is over the softcap
       return this.attributeScalingLimit + (this.attributeScalingLimit * 9 / 4) +
         (this.attributeScalingLimit * 90 / 20) +
-        (this.attributeScalingLimit * (this.attributeSoftCap - 100) / 100) +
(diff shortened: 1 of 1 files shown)


Croebh:
Oof, that is a significant 'nerf' to aptitude. 

That takes my current aptitude multiplier from 322,187.45 to 3,187.45, so roughly 1% of what it was.

immortalityidle:
We should probably do this. But I think when we introduce this we also need to introduce something else that lets them get equivalent gains. Otherwise the whole endgame balance goes all off.

Hornsoup:
I feel like making it longer to get to those later numbers is important, but I'll have to review the math design of it later cause I'm taking time for myself instead of allowing the hyperfocus to seat itself deeper.

Hornsoup:
I updated the softcap to cover for this, it's in another pull and it includes this fix, though the formula is gentle enough to stand alone.

PR #40, opened 2022-07-18
fix: Allow auto merge to equipped with single auto equip manual

PotatoArt:
Need to address both TODOs before this should be merged. 

Right now, autoequipbest runs even if everything is going to get merged, though it only happens on long ticks, it sweeps through the entire inventory and checks if it's null, and if not, checks if it's equipment. Nothing will be, but these are checks that can be short circuited if everything is getting merged to equipped slots.

Secondly, the autoequipBestEnabled option isn't consulted at all once merging-to-equipment is unlocked. Either the option shouldn't be available, or the option should do something.


src/app/game-state/inventory.service.ts
```diff
@@ -1122,16 +1122,19 @@ export class InventoryService {
...
-    if (destinationItem != null && this.autoequipBestWeapon && this.autoequipBestArmor){
+    //TODO: We might want to check if auto-equip-best is enabled at this point, or remove the tickbox
+    // finally, merge the last item with that slot into the equipped item (if present and corresponding autoequip is unlocked)
+    if (destinationItem != null && (this.autoequipBestWeapon || this.autoequipBestArmor)){
       sourceItem = this.characterService.characterState.equipment[slot];
       if (sourceItem == null){
         return;
       }
-      if (slot == 'rightHand' || slot == 'leftHand'){
+      if (slot == 'rightHand' || slot == 'leftHand' && this.autoequipBestWeapon) {
```
ChandlerFerry, line 1132:
```suggestion
      if ((slot == 'rightHand' || slot == 'leftHand') && this.autoequipBestWeapon) {
```

ChandlerFerry:
@PotatoArt I presume this is still a work in progress?

PotatoArt:
Oh. What I meant was "before you merge this, let me know which option we should go with on these two". The first point I made it so it checks. The second point, I leaved it as is, where it ignores the option, but I strongly feel that either it shouldn't exist anymore, or it should be hidden. I don't really care which, so I was looking for feedback/opinions.

In either case, there's a conflict now, so I guess it's not ready anymore.

immortalityidle (approved):

PR #42, opened 2022-07-18
Balanced softcap around fix.

Hornsoup:
Gains are now meaningful after softcap too, I'd even recommend removing all the extra code and just using this formula because the curve is so nice and fits right in. Also fixed spirituality lifespan to gain from aptitude instead of value.

### DIFF
--- a/src/app/game-state/character.ts
+++ b/src/app/game-state/character.ts
@@ -324,7 +324,7 @@ export class Character {
     if (this.money > this.maxMoney){
       this.money = this.maxMoney;
     }
-    this.spiritualityLifespan = this.getAptitudeMultipier(this.attributes.spirituality.value) * 5;
+    this.spiritualityLifespan = this.getAptitudeMultipier(this.attributes.spirituality.aptitude) * 5;
     this.lifespan = this.baseLifespan + this.foodLifespan + this.alchemyLifespan + this.statLifespan + this.spiritualityLifespan + this.magicLifespan;
     this.accuracy = 1 - Math.exp(0 - this.getAptitudeMultipier(this.attributes.speed.value) * this.accuracyExponentMultiplier);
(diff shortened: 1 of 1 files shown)


Hornsoup:
Yes, if done from the start this curve makes sense because it would take quite a long time to get that much aptitude after hitting the softcap because it slows down the ascent to 10 billion charisma, and if you're at Death then you're nearly ageless anyway, so having extreme time to kill Death isn't the point of that fight. The curve can be lowered if it's too high though, just need to increase the power by tiny amounts, probably keep between 0.5 and 0.6.

kagejohn:
I have used most followers, and i try my best to train them as high as possible

Hornsoup:
I just looked at some people reporting current multipliers of 5 million in the live build, which shouldn't be possible. The multiplier should be heavily softcapped at 10k and so now I'm quite confused.

kagejohn:
Condense Soul Core, lowers aptitudeGainDivider by 10 for each rank.
At rank 9 that gives a multiplier of just over 5 million

kagejohn:
For Reinforce Meridians the ranks give these limits:
0: ~10k, 1: ~20k, 2: ~40k, 3: ~80k, 4: ~161k, 5: ~322k, 6: ~644k, 7: ~1288k, 8: ~2577k, 9: ~5154k

Hornsoup:
> For Reinforce Meridians the ranks give these limits:
> 0: ~10k, 1: ~20k, 2: ~40k, 3: ~80k, 4: ~161k, 5: ~322k, 6: ~644k, 7: ~1288k, 8: ~2577k, 9: ~5154k

No, they should give 0: 10 etc. these are way off from what they should be, a bug slipped in somewhere on live.

immortalityidle:
Manually merged.

PR #44, opened 2022-07-18
Soul core meaningful

Hornsoup:
Makes for more meaningful Soul core by keeping up to 9% aptitude on ascension. <strike>Predicated on the aptitude fix.</strike>


src/app/game-state/character.ts
```diff
@@ -374,8 +374,8 @@ export class Character {
       // increase by log2 of whatever is over the softcap
```
immortalityidle, line 374: Can you update this comment? I'm not good enough at math to get what this is doing right away.

Hornsoup: Oop, yeah that's part of the aptitude fix, forgot the comment there. Also added a local variable for jittering the balance as needed.

Hornsoup:
This will likely need updating after moving this code to a method but it's been de-branched.

ChandlerFerry (approved):
Code looks good, but balance is not something I can weigh in on.

immortalityidle (approved):

PR #45, opened 2022-07-18
chore: nvmrc init

ChandlerFerry:
As someone who developers in v16 Active LTS daily, swapping to this repo throws me a bit off course without an `.nvmrc`.


.nvmrc
```diff
@@ -0,0 +1 @@
+v12
```
ChandlerFerry, line 1: If this repository is not intended to be `v12`, please change as you see fit and update `@types/node` to follow suite.

PR #46, opened 2022-07-18
Auto-Buyer Settings Update

Kleptichron:
This adds the following table to settings:
image
This table allows for shifting priorities of the various auto-buyers by clicking the buttons on the left, and can also allow behavior changes through the checkboxes on the right. 

"Enabled" means that the auto-buyer is running. Currently there's no way to "turn off" some auto-buyers, like furniture, so this allows for that.

"Wait for Finish" means that we should finish automatically upgrading an item before moving on to the next priority. For example, complete housing upgrades before moving on to land/fields. In cases where a buyer is blocked (we have to upgrade housing before we can buy all furniture), this setting will be ignored.

This moves auto-buying services out of the home service, except for settings that already existed in home service, since I didn't want to break saves. :)

I also added a manual for this, because progressive unlocking was hard to figure out. Before unlocking this manual, autobuyers will default to a behavior _close_ to what's currently done: Home -> Furniture -> Land, all enabled, all waiting for the previous to finish. These will run every tick, but will immediately skip if something isn't unlocked.

The manual unlocks when all parts of the customization will make sense: home autobuy, land autobuy, field autobuy, and furniture autobuy.

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -278,7 +278,7 @@ export class AchievementService {
     {
       name: "Industrial Revolution",
       description: "You've found all the basic autobuyers and unlocked the " + this.itemRepoService.items['autoBuyerSettingsManual'].name,
-      hint: "Immortals have discerning taste in furnishings.",
+      hint: "Become really, really lazy",
       check: () => {
         return this.homeService.autoBuyHomeUnlocked &&
           this.homeService.autoBuyLandUnlocked &&


src/app/game-state/home.service.ts
```diff
@@ -735,103 +734,14 @@ export class HomeService {
 
   }
 
-  autoBuy(){
```
Kleptichron, line 738: Probably biggest change here. It seemed like autobuy could be moved to its own service, especially if there was eventually a bought resource that didn't have anything to do with the home!

src/app/game-state/achievement.service.ts
```diff
@@ -275,6 +275,21 @@ export class AchievementService {
       },
       unlocked: false
     },
+    {
```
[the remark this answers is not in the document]
Kleptichron: Crap that hint. I'll fix it.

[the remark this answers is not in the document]
Kleptichron: I would _never_ just copy and paste something badly...

src/app/game-state/item-repo.service.ts
```diff
@@ -1167,6 +1169,29 @@ export class ItemRepoService {
...
+          this.autoBuyerService = this.injector.get(AutoBuyerService);
+        }
+        this.autoBuyerService.autoBuyerSettingsUnlocked = true;
+        this.logService.addLogMessage("The teachings of the manual sink deep into your soul. You'll be able to apply this knowledge in all future reincarnations.", "STANDARD", 'EVENT');
+      },
+      owned: () => {
+        if (!this.autoBuyerService){
+          this.autoBuyerService = this.injector.get(AutoBuyerService);
+        }
+        return this.autoBuyerService.autoBuyerSettingsUnlocked;
```
immortalityidle, line 1192: Are we preserving this in the saved settings?

Kleptichron: It has a the get/set properties methods and it's added to gameState. So I think so!

Kleptichron:
@immortalityidle Yep, Wait for Finish will handle that. The "house" auto-buyer still buys land in order to upgrade houses, so if you use the settings I screenshot above, it will do this:

https://user-images.githubusercontent.com/109362801/179604294-b1cb795a-502b-4f96-aaa0-99578f32ecaf.mov

PR #47, opened 2022-07-18
Quick softcap fix for Meridian ranks

Hornsoup:


Hornsoup:
It's safer to slow down progress than to speed it up, so taking the safer measure by increasing the divisor power for now.

immortalityidle (approved):

PR #51, opened 2022-07-19
Rebalances Empowerment

Hornsoup:
Changes the curvature of empowerment but not the time it takes to achieve max. Currently about a month of pure grinding to max.

Also adds Ascension requirement to add the pills consumed to the factor.


immortalityidle:
I don't mind changing the curve for the empowerment effect, but I don't think we need the rest of this. I like the empowerment pills being a simple buff that persist across reincarnations/ascensions.

Hornsoup:
Okay, I'll revert all of the extra and make a new PR, but what do you think about balancing the pills around x10 since the previous multiplier (before the 5 million bug) was around 10k? Balance it to take about a day's grind to make 1000 pills, and if people want to go nuts for million multipliers that's up to them?

immortalityidle:
Heh, love your optimism that people will be playing this for the next three months. I'll defer to your judgement on the math curve for them, as long as players can see decent gains for them early on.

Hornsoup:
It won't be quite what people *want* but 1 day for a 10x mult is already alot higher than the game has been balanced for thus far. I'll work something in around that then.

PR #54, opened 2022-07-19
Fix: Improving home information

PotatoArt:
Empowerment pill clarification can be split out if this tooltip moving isn't appreciated.

House tooltips needed to be updated, since "a bit of health" is inaccurate when it recovers 80, and it doesn't mention mana recovery at all. While we're at it, might as well have the upgrade tooltip say what the advantage of an upgraded house will be. 

Possible changes to consider: 
- changing from "next house's values" to "difference between this house and next house". 
- moving all of this back to home-panel's html and figuring out how to put if statements in html. 
- creating and using variables in the descriptions to make possible/eventual globalization and localization marginally easier.
  - if variables are created, use them inside of consequence function.

### DIFF
--- a/src/app/ascension-store-modal/ascension-store-modal.component.html
+++ b/src/app/ascension-store-modal/ascension-store-modal.component.html
@@ -55,6 +55,6 @@
     </tr>
   </table>
   <span *ngIf="characterService.characterState.empowermentFactor > 1">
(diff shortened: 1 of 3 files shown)


immortalityidle:
This looks good at first glance. I'll give it a proper review later today. Are you on the discord so I can give you a contributor role?

immortalityidle:
Ack! didn't mean to close it.

immortalityidle (approved):

PR #58, opened 2022-07-20
Improve contrast of active time panel button

tobbez:
Because the contrast between the 'active' background color and regular
background color is so low, telling which button in the time panel is
active is not trivial (especially in the dark mode).

Swapping the foreground and background colors of the active button makes
it easy to tell it apart from the others at a glance, while preserving
the color theme.

### DIFF
--- a/src/app/time-panel/time-panel.component.less
+++ b/src/app/time-panel/time-panel.component.less
@@ -42,7 +42,8 @@ label {
 }
 
 .iconActive {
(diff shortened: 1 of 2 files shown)


immortalityidle:
Mind posting screenshots for easy reference? Dark and light please?

tobbez:
image
image

immortalityidle (approved):
Looks good. Ding me on the discord if you want a contributor role there.

PR #59, opened 2022-07-20
Empowerment Balance try 2

Hornsoup:
People have already hit cap. Too late to take it back and too late to balance but this exists.

### DIFF
--- a/src/app/ascension-store-modal/ascension-store-modal.component.html
+++ b/src/app/ascension-store-modal/ascension-store-modal.component.html
@@ -55,6 +55,6 @@
     </tr>
   </table>
   <span *ngIf="characterService.characterState.empowermentFactor > 1">
-    You have consumed {{(characterService.characterState.empowermentFactor - 1) * 100 | number:'1.0-0'}} Empowerment Pills which multiplies all your attribute gains by {{characterService.characterState.empowermentFactor  | number:'1.0-2'}}. You will always keep this bonus, even after ascending.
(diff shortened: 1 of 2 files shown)


immortalityidle:
Not sure what this math does, but it looks fancy. We should probably make a corresponding change to where we show the value in the ascension panel too though.

Hornsoup:
Math just sets x10 to about a day and shoots the rest out to the sky so people can prove me wrong in how quickly they can farm it with the new time dilation. I'm still kinda working on bits, but I'll also add in the resultant value in ascension alongside pill count. I also reduced cap to 100 because it was just insane to have people with a 10k stat mult after grinding spirituality for a bit and dilating time with their insane age, it compounds.

Hornsoup:
It's untested due to time constraints but the math checks out.

ChandlerFerry:
Personal opinion is that this game needs a balance sweep as a whole as it's difficult to commit to playing with everything fluctuating so heavily.

The numbers just don't quite make sense in a lot of cases, but I don't have the necessary experience to say exactly what they should be.

ChandlerFerry (approved):
Code looks good.

Balance is TBD by someone else.

Hornsoup:
Problem with balance is that there was none on release, it wasn't tested end-to-end so at the start Christian was patching for people to get through content. The early game is solid, but it's around 4/1/3 that things got bad and every fix thereafter has been damage control, sometimes damage.

PR #62, opened 2022-07-21
Added ctrl and shift to follower dismiss

Hornsoup:
Shift does the old removing everyone and setting to 0, whereas ctrl doesn't remove anyone and sets the current number to the target so nobody new is hired.


src/app/game-state/followers.service.ts
```diff
@@ -289,9 +289,32 @@ export class FollowersService {
...
 
-  dismissFollower(follower: Follower){
-    const index = this.followers.indexOf(follower);
-    this.followers.splice(index, 1);
+  /**
+   * 
+   * @param follower the Follower interface of the selected follower.
+   * @param option 1 to dismiss selected follower, 0 to dismiss the job, -1 to limit number of the selected job to current amount employed.
+   */
+  dismissFollower(follower: Follower, option = 1){
```
immortalityidle, line 297: Feels like a weird use of parameters here. It's essentially three independent functions with no overlap split by the if on the option. Might make for more readable code to just break it into three.

Hornsoup: Methods made.

immortalityidle (approved):

PR #66, opened 2022-07-21
feature: Auto pause customization

PotatoArt:
So, this is mostly being created so that people can look over what I have and discuss it. If I didn't mess anything up (which would need testing to confirm) this shouldn't have any effect besides defaulting pauseOnDeath to true add an achievement and manual that do nothing. 

This is basically an overengineered "pause before death" that only cares about deaths of natural causes. Hoping this is engineered enough to allow adding of any other reasons people might want to pause. Say, having 0 money, or very low health, or getting knocked out. Might need to add internal cooldowns for those, like how TimeAutoPauser works.

Need to add the actual options to the options modal before people can tick anything but pauseOnDeath on, but I'm looking for ideas on things like achievement condition, name, manual cost, hint, etc. All the cool ideas I have would need to add a tracker (times paused, etc). And anyone who sees a problem with how this is currently set up, let me know.

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -109,6 +109,18 @@ export class AchievementService {
       unlocked: false
     },
     {
+      name: "Won't Live Forever",//TODO update description and requirements
+      description: "PLACEHOLDER DESCRIPTION " + this.itemRepoService.items['autoPauseSettingsManual'].name,
+      hint: "PLACEHOLDER HINT",
+      check: () => {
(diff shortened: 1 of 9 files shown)


immortalityidle:
Haha, I just put in something very similar, though with a significantly different implementation. If you're going to work on something, please put it in the doing column on the Trello so we don't duplicate effort (if you don't have edit access there pm me your email address on discord and I'll get you added). Take a look at my changes and see where you want to go from there, then let's discuss in discord.

PotatoArt:
no reason to leave this open

PR #69, opened 2022-07-21
Added fix to reincarnation for impossible tasks

Hornsoup:
Code looks good, but still working on testing.

### DIFF
--- a/src/app/game-state/impossibleTask.service.ts
+++ b/src/app/game-state/impossibleTask.service.ts
@@ -169,21 +169,29 @@ export class ImpossibleTaskService {
     });
   }
 
-  reset(){
-    let completeTaskIndex = -1
-    for (let taskIndex = 0; taskIndex < this.tasks.length; taskIndex++){
-      if (this.taskProgress[taskIndex].complete){
-        completeTaskIndex = taskIndex; // Sanity check for the latest complete task
+  markPriorCompletions(){
+    for (let taskIndex = this.tasks.length - 1; taskIndex >= 0; taskIndex--){
+      if (this.taskProgress[taskIndex].complete){ // Sanity check for the latest complete task
+        this.nextTask = taskIndex + 1;
+        while(taskIndex >= 0){
+          this.taskProgress[taskIndex].progress = this.tasks[taskIndex].progressRequired;
+          this.taskProgress[taskIndex].complete = true;
+          taskIndex--
+        }
+      } else if (taskIndex === 0){
+        this.nextTask = 0;
       }
     }
-    for (let taskIndex = 0; taskIndex < this.tasks.length; taskIndex++){
-      if (taskIndex < completeTaskIndex){
-        this.taskProgress[taskIndex].complete = true;
-      }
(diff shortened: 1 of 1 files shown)


src/app/game-state/impossibleTask.service.ts
```diff
@@ -170,7 +170,16 @@ export class ImpossibleTaskService {
...
 
   reset(){
+    let completeTaskIndex = -1
+    for (let taskIndex = 0; taskIndex < this.tasks.length + 1; taskIndex++){
+      if (this.taskProgress[taskIndex].complete){
+        completeTaskIndex = taskIndex; // Sanity check for the latest complete task
+      }
+    }
     for (let taskIndex = 0; taskIndex < this.tasks.length; taskIndex++){
+      if (taskIndex < completeTaskIndex){
```
immortalityidle, line 180: Thought: If we loop through backwards we could do this in one loop.

Hornsoup: Thought in reverse, I should have. I'll update.

Hornsoup: Oop, and testing again with loading an old save caused an issue, time to find and fix.

src/app/game-state/impossibleTask.service.ts
```diff
@@ -169,12 +169,29 @@ export class ImpossibleTaskService {
     });
   }
 
+  markPriorCompletions(){
+    for (let taskIndex = this.tasks.length - 1; taskIndex >= 0; taskIndex--){
```
ChandlerFerry, line 173: To my knowledge, `--` is ~30% slower in the V8 engine. Please ensure the tradeoff is worthwhile.

src/app/game-state/impossibleTask.service.ts
```diff
@@ -169,12 +169,29 @@ export class ImpossibleTaskService {
...
+      if (this.taskProgress[taskIndex].complete){ // Sanity check for the latest complete task
+        this.nextTask = taskIndex + 1;
+        while(taskIndex >= 0){
```
immortalityidle, line 176: I don't think you need this extra loop, since you'll be hitting the rest of them in the outer loop. Just set a boolean flag that you've found one and then set complete =true if that found flag is set.

src/app/game-state/impossibleTask.service.ts
ChandlerFerry: After some testing, this knowledge might be dated.

src/app/game-state/impossibleTask.service.ts
Hornsoup: That ignores the setting of this.nextTask above that loop, which is the primary issue, and I'd have to do a secondary loop to check values for that if I do as suggested. The inner loop allows me to set the nextTask at that point without having to do a bunch of other stuff and finishes out the outer loop.

Hornsoup: Wait no, you're right, I'm reading this wrong.

Hornsoup:
Should be good, the tests I could think of came back clean.

immortalityidle (approved):

PR #72, opened 2022-07-21
Some options fixes while I balance

Hornsoup:
If the numerical options fields are left blank it causes issues. Simple fix.

### DIFF
--- a/src/app/game-state/character.ts
+++ b/src/app/game-state/character.ts
@@ -356,7 +356,8 @@ export class Character {
 
   getEmpowermentMult(): number{
     const max = 99;
-    return 1 + 2 * max / (1 + Math.pow(1.02, (-this.empowermentFactor / 3))) - max;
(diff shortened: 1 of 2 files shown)


src/app/game-state/character.ts
```diff
@@ -356,7 +356,8 @@ export class Character {
 
   getEmpowermentMult(): number{
     const max = 99;
-    return 1 + 2 * max / (1 + Math.pow(1.02, (-this.empowermentFactor / 3))) - max;
+    const empowermentFactor = this.empowermentFactor - 1;
+    return 1 + 2 * max / (1 + Math.pow(1.02, (-empowermentFactor / 3))) - max;
```
Hornsoup, line 360: This wasn't returning 1 like it should have been because I didn't set x (empowermentFactor) to 0 since it defaults to 1.

immortalityidle (approved):

PR #75, opened 2022-07-22
Aptitude multiplier balancing

cqkok:
Balance update:
- Aptitude multiplier will grow very slow once more than 100,000, thus fixing the problem of too many lifespan and too many stats

Reducing the cost to train followers to become 500m
and reducing the cost of bloodline 5 to become 9.375T
in order to make sure the game not become too hard after the nerf.

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -431,6 +431,24 @@ export class AchievementService {
       unlocked: false
     },
     {
+      name: "Unlimited Taels",
+      description: "You have unlocked bloodline 4 and now you have unlimited money to spent",
+      hint: "Rumour said that some people are too rich.",
+      check: () => {
+        return this.characterState.bloodlineRank >= 4;
+      },
+      effect: () => {
+        if (this.characterState.bloodlineRank == 4) {
+          this.characterState.bloodlineCost = 10000000000000
+        }
+        
+        if (this.characterState.bloodlineRank == 5) {
+          this.characterState.bloodlineCost = 1000000000000000
+        }
+      },
+      unlocked: false
+    },
+    {
       name: "Grandpa's Old Tent",
       description: "You've gone through eight cycles of reincarnation and come to understand the value of grandfathers.",
       hint: "Just keep playing. I'm sure this will come to an aspiring immortal eventually.",
--- a/src/app/game-state/character.service.ts
+++ b/src/app/game-state/character.service.ts
@@ -232,7 +232,12 @@ export class CharacterService {
     this.logService.addLogMessage(
       "You will be reborn into your own family line and reap greater benefits from your previous lives.",
       'STANDARD', 'STORY');
-    this.characterState.bloodlineCost *= 1000;
+    // Bloodline 1 Cost: base cost
+    // Bloodline 2 Cost: 1,000 x base cost
+    // Bloodline 3 Cost: 750,000 x base cost
+    // Bloodline 4 Cost: 375,000,000 x base cost
+    // Bloodline 5 Cost: 93,750,000,000 x base cost
+    this.characterState.bloodlineCost *= 1000 - 250 * this.characterState.bloodlineRank;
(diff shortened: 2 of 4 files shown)


cqkok:
Bloodline 5 spirituality cost need reworks... I forgot that if someone load the game with saves then their bloodlineCost will be wrong

cqkok:
Used some weird way to change the bloodline 5 spiritual cost

src/app/game-state/achievement.service.ts
```diff
@@ -430,6 +430,24 @@ export class AchievementService {
...
+      hint: "Rumour said that some people are too rich.",
+      check: () => {
+        return this.characterState.bloodlineRank >= 4;
+      },
+      effect: () => {
+        if (this.characterState.bloodlineRank == 4) {
+          this.characterState.bloodlineCost = 10000000000000
+        }
+        
+        if (this.characterState.bloodlineRank == 5) {
```
immortalityidle, line 445: Not sure this will work. I don't think we have anything that guarantees this will trigger when we hit bloodline 5. Probably better to put this logic where the bloodline cost increases in upgradeBloodline.

src/app/game-state/character.service.ts
```diff
@@ -234,6 +234,9 @@ export class CharacterService {
       'STANDARD', 'STORY');
     this.characterState.bloodlineCost *= 1000;
     this.characterState.bloodlineRank++;
+    if (this.characterState.bloodlineRank >= 4) {
```
immortalityidle, line 237: Hmm. The logic is here. Why do we also have it in the achievement?

cqkok: I found out that this function only being called when player upgrade their bloodline.

But what if a player originally already have bloodline 4 or 5, and the bloodlinecost already stored in their save file?

The achievement is trying to change the cost saved in their save file

cqkok: I not sure how the save file works, whether the save file save the bloodlinecost? or it save the bloodlineRank and reupgrade it back on loading? If it is saving the bloodlineRank and reupgrade when loading, then this function will make sure the bloodlineCost is correct

cqkok: my idea is, if the bloodline cost is saved in the save file, then when they load the save files, the new achievement will immediately pass the "check" function, so the effect will be applied to change the cost.

A very bad idea anyway, achievement should not be use like that

immortalityidle: Yeah. If you look in character.ts you can see the cost and the rank in the CharacterProperties. Anything in that interface gets saved into the save file.

cqkok: thats the reason why i add an achievement to make "correction" to the saved bloodline cost, because the old bloodline cost won't be affected by the change in the upgradeBloodline function

immortalityidle: Seems like we could just set the right cost here and be done. Whatever we set it to should be persisted in the save.

cqkok: The problem is, this function only runs when the player have enough spirit.

If the player don't have enough spirit, then they will be blocked at the store.service

So let's say a player done bloodline 4 before patch

So inside this function (before patch), the function will set bloodline 5 price to become 100T, and then lv up the bloodline rank from 3 to become 4.

The 100T is saved inside the save file.

So after we patch it, the player will still see 100T cost in the UI, when the try to click, they will also get a failed message from the store.service, and this function won't be get called.

-------------

Player who originally have bloodline  3 before we patch, will be normal. But player who originally have bloodline 4 & 5, will have the old 100T and 100q stored in their save file.

What my achievement try to do is to make sure that all those player who have bloodline 4 / 5 before patch, will get their value updated in the achievement effect.

immortalityidle: Ah, yeah, I see. Forgot about legacy save files. My bad.

immortalityidle (approved):

PR #76, opened 2022-07-22
All attribute gain is now more consistent

cqkok:
This greatly reduces the impact of luck affecting player's stat

### DIFF
--- a/src/app/game-state/activity.service.ts
+++ b/src/app/game-state/activity.service.ts
@@ -993,11 +993,11 @@ export class ActivityService {
       consequenceDescription:
         ['Uses 5 Stamina. Increases a random attribute and provides a little money.'],
       consequence: [() => {
-        const keys = Object.keys(
-          this.characterService.characterState.attributes
-        ) as AttributeType[];
-        // raise all first five stats
-        keys.forEach(key => this.characterService.characterState.increaseAttribute(key, 0.02));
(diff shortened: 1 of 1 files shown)


src/app/game-state/activity.service.ts
```diff
@@ -996,9 +996,8 @@ export class ActivityService {
         const keys = Object.keys(
           this.characterService.characterState.attributes
         ) as AttributeType[];
-        // randomly choose any of the first five stats
-        const key = keys[Math.floor(Math.random() * 5)];
-        this.characterService.characterState.increaseAttribute(key, 0.1);
+        // raise all first five stats
+        keys.forEach(key => this.characterService.characterState.increaseAttribute(key, 0.02));
```
kagejohn, line 1000: Wouldn't this raise every stat by 0.02 and not just the first 5?

cqkok: I not too sure what is insides "keys"...

But from the line of code I removed, I saw that the "key" is generated by randomly pick 1 from within the "keys"

So I guess what inside the "keys" array is the 5 main attribute

cqkok: you are right. this code will raise every stat because the "keys" array contains every stat

immortalityidle (approved):

PR #77, opened 2022-07-22
Create a new achievement

cqkok:


immortalityidle (approved):
Don't be offended if I tweak the text before release, but I like the idea.

PR #83, opened 2022-07-23
Suggestion for the usage of Monster Gem

cqkok:


src/app/game-state/activity.service.ts
```diff
@@ -1920,6 +1923,71 @@ export class ActivityService {
       unlocked: false,
       skipApprenticeshipLevel: 0
     };
+    
+    this.CombatCultivation = {
+      level: 0,
+      name: ['Combat Cultivation'],
+      activityType: ActivityType.CombatCultivtion,
+      description: ['Focus on the development of your combat skill.'],
+      consequenceDescription: ['Uses 200 Stamina. A very advanced cultivation technique. Make sure you have killed some monster before attempt this, teaches you how to do combat more effectively.'],
```
Auchtor, line 1932: `      consequenceDescription: ['Uses 200 Stamina. A very advanced cultivation technique. Make sure you have killed some monsters before attempt this, teaches you how to do combat more effectively.']`

You may want to consider using the plural of monster here.

immortalityidle:
I like the idea. I was actually thinking of something along these lines while I was away this weekend. Notes: since this just upgrades gear I think we should change the name to something along the lines of "Infuse Equipment". I also think it should also have a mana cost with it, scaling up with the power of the increase. Not sure I love the way it does the weapons then maybe does the armor, I think I'd prefer it either hit a random slot or have an option that lets you choose what to upgrade.

src/app/game-state/activity.service.ts
```diff
@@ -1920,6 +1923,71 @@ export class ActivityService {
...
+      consequence: [() => {
+        this.characterService.characterState.status.stamina.value -= 200;
+        
+        const gemGrade = this.inventoryService.consume('spiritGem') / 10;
+
+        if (gemGrade > 0) {
+          this.battleService.tickCounter += 10;
```
immortalityidle, line 1939: I don't know if this belongs with the rest. I'm not opposed to having it as an activity, but I think it should be an activity of its own.

src/app/game-state/activity.service.ts
cqkok: I have applied the changes in new commit

src/app/game-state/activity.service.ts
cqkok: my original intention is since this activity can be run every day with high building such as capital, so i add tick counter to make sure there is enough gem to be consumed.

I removed it and also make this activity now with mana cost. Feel free to edit the values, I am just trying to think what can be done with the monster gem according to your trello's to do.

immortalityidle (approved):

PR #86, opened 2022-07-23
Re-balance and various fixes

Hornsoup:
Please see notes.

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -159,7 +159,7 @@ export class AchievementService {
       check: () => {
         return this.characterService.characterState.empowermentFactor > 1;
       },
-      effect: () => {
+      effect: () => { //TODO: Create a downside to taking empowerment pills, maybe post-Death
       },
       unlocked: false
     },
@@ -443,19 +443,12 @@ export class AchievementService {
     },
     {
       name: "Unlimited Taels",
-      description: "You have unlocked bloodline 4 and now you have unlimited money to spent",
-      hint: "Rumour said that some people are too rich.",
+      description: "Your family has unlocked the secrets of exploiting capital. You probably never have to worry about money again.",
+      hint: "You don't want to stop the exploitation, you want to become the exploiter.",
       check: () => {
         return this.characterService.characterState.bloodlineRank >= 4;
       },
       effect: () => {
-        if (this.characterService.characterState.bloodlineRank == 4) {
-          this.characterService.characterState.bloodlineCost = 10000000000000
-        }
-        
-        if (this.characterService.characterState.bloodlineRank == 5) {
-          this.characterService.characterState.bloodlineCost = 1000000000000000
-        }
       },
       unlocked: false
     },
@@ -589,10 +582,10 @@ export class AchievementService {
     {
       name: "Impossible",
       description: "You have achieved incredible power and are ready to begin taking on impossible tasks.",
-      hint: "No one can achieve this. It can't be done.",
+      hint: "No one can reach the limits of humanity. It can't be done.",
       check: () => {
-        return (this.characterService.soulCoreRank() >= 4) &&
-          (this.characterService.meridianRank() >= 4) &&
+        return (this.characterService.soulCoreRank() >= 9) &&
+          (this.characterService.meridianRank() >= 9) &&
           this.characterService.characterState.bloodlineRank >= 5;
       },
       effect: () => {
@@ -650,6 +643,23 @@ export class AchievementService {
       },
       effect: () => {
         this.storeService.unlockManual(this.itemRepoService.items['followerAutoDismissManual']);
+      },
+      unlocked: false
+    },
+    {
+      name: "Loyal Followers",
+      description: "One of your followers has trained under you so long they have nothing else to learn. In an epiphany you realized how to double your new followers' lifespan.",
+      hint: "Endless training.",
+      check: () => {
+        for (const follower of this.followerService.followers){
+          if (follower.power === 100){
+            return true;
+          }
+        }
+        return false;
+      },
+      effect: () => {
+        this.followerService.followerLifespanDoubled = true;
       },
       unlocked: false
     },
--- a/src/app/game-state/character.ts
+++ b/src/app/game-state/character.ts
@@ -80,12 +80,12 @@ export class Character {
   dead = false;
   attributeScalingLimit = 10;
   attributeSoftCap = 100000;
-  aptitudeGainDivider = 100;
+  aptitudeGainDivider = 5 * Math.pow(1.5, 9); // Exponential Soul Core ranks, up to 20%
   condenseSoulCoreCost = 10;
   condenseSoulCoreOriginalCost = 10;
   reinforceMeridiansCost = 1000;
   reinforceMeridiansOriginalCost = 1000;
-  bloodlineCost = 100;
+  bloodlineCost = 1000;
   bloodlineRank = 0;
   manaUnlocked = false;
   accuracy = 0;
@@ -258,6 +258,8 @@ export class Character {
     this.statLifespan = this.getAptitudeMultipier(totalAptitude / 5);
     if (this.bloodlineRank < 5){
       this.statLifespan *= 0.1;
+    } else {
+      this.statLifespan *= 5;
     }
 
     const keys = Object.keys(this.attributes) as AttributeType[];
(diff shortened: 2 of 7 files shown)


src/app/game-state/character.ts
```diff
@@ -498,13 +496,14 @@ export class Character {
...
-    this.aptitudeGainDivider = properties.aptitudeGainDivider;
     this.condenseSoulCoreCost = properties.condenseSoulCoreCost;
+    // This is derived to avoid save issues. Calculate rank and subtract from power to reduce the exponential aptitude divider.
+    this.aptitudeGainDivider = 5 * Math.pow(1.5, 9 - Math.log10(this.condenseSoulCoreCost / this.condenseSoulCoreOriginalCost)); 
     this.reinforceMeridiansCost = properties.reinforceMeridiansCost;
-    this.bloodlineCost = properties.bloodlineCost;
+    this.attributeScalingLimit = properties.attributeScalingLimit;
+    this.attributeSoftCap = properties.attributeSoftCap;
     this.bloodlineRank = properties.bloodlineRank;
+    this.bloodlineCost = 1000 * Math.pow(100, this.bloodlineRank); // This is derived to avoid save issues.
```
Hornsoup, line 505: Moved a lot around and derived values during the setProperties to avoid save issues. I recommend saving ranks for all three Ascensions rather than costs, but that might be hard to backcompat and it's not worth it at this point.

src/app/game-state/followers.service.ts
```diff
@@ -273,12 +277,13 @@ export class FollowersService {
       this.logService.addLogMessage("A new follower shows up, but they were a " + job + " and you don't want any more of those.","STANDARD","FOLLOWER");
       return;
     }
-
-    this.logService.addLogMessage("A new follower has come to learn at your feet.","STANDARD","FOLLOWER");
+    
+    const lifespanDivider = this.followerLifespanDoubled ? 5 : 10;
```
Hornsoup, line 281: Added in follower lifespan doubling as an achievement reward for getting a follower to level 100. If a player can't do that then they can't continue with swimming. It's an intentional wall.

src/app/game-state/followers.service.ts
```diff
@@ -273,12 +277,13 @@ export class FollowersService {
...
+    this.logService.addLogMessage("A new " + job + " has come to learn at your feet.","STANDARD","EVENT");
     this.followers.push({
       name: this.generateFollowerName(),
       age: 0,
-      lifespan: this.characterService.characterState.lifespan,
+      lifespan: this.characterService.characterState.lifespan / lifespanDivider,
```
Hornsoup, line 286: Not only reverted this overpowered change, but got an idea to create the achievement.

src/app/game-state/impossibleTask.service.ts
```diff
@@ -51,7 +51,7 @@ export class ImpossibleTaskService {
       name: "Swim to the bottom of the ocean",
       description: "You find a scrap in an ancient text that leads you to believe that the secret of immortality lies buried deep beneath the ocean's currents.",
       taskType: ImpossibleTaskType.Swim,
-      progressRequired: 100000,
+      progressRequired: 400000,
```
Hornsoup, line 54: Increased progress required alongside follower life and attributes giving more, player has just enough follower lifespan for a hunter or two to keep player alive.

src/app/game-state/store.service.ts
```diff
@@ -129,7 +129,7 @@ export class StoreService {
       this.bloodLineHomeRequirement = this.homeService.homesList[HomeType.Fortress];
     } else if (this.characterService.characterState.bloodlineRank === 4){
       // Basic Stat Lifespan
-      this.bloodlineDescription = "End your current life, sacrifice all attributes and aptitudes, and enhance your bloodline. Your armor and your weapons equipped on death will become family heirlooms and will be inherited by your future self. You will also inherit your past self's money plus interest. Your aptitudes & spirituality extend your lifespan to a much greater degree.";
+      this.bloodlineDescription = "End your current life, sacrifice all attributes and aptitudes, and enhance your bloodline. Your armor and your weapons equipped on death will become family heirlooms and will be inherited by your future self. You will also inherit your past self's money plus interest. Your aptitudes extend your lifespan to a much greater degree.";
```
Hornsoup, line 132: Removed the spirituality lifespan increase in bloodline and replaced with a stat lifespan increase.

Hornsoup (commented):
See review points for changes in the first round up to Swimming.

src/app/game-state/followers.service.ts
```diff
@@ -248,12 +251,13 @@ export class FollowersService {
...
     this.autoDismissUnlocked = properties.autoDismissUnlocked || false;
     this.maxFollowerByType = properties.maxFollowerByType || {};
+    this.followerLifespanDoubled = properties.followerLifespanDoubled || false;
   }
 
   generateFollower(){
     this.followersRecruited++;
     if (this.followers.length >= this.followerCap){
-      this.logService.addLogMessage("A new follower shows up, but you already have too many. You are forced to turn them away.","INJURY","FOLLOWER");
+      this.logService.addLogMessage("A new follower shows up, but you already have too many. You are forced to turn them away.","INJURY","EVENT");
```
kagejohn, line 260: I don't understand why you made this change, wasn't the idea that you should be able to hide these unnecessary logs?
At the stage of the game i am at you get one of these every ~4 seconds.

Hornsoup:
This change is because it's a reaction to player input, since there should
be feedback for the activity, especially when a limit is hit or the action
succeeds.

src/app/game-state/followers.service.ts
Hornsoup: This change is because it's a reaction to player input, since there should be feedback for the activity, especially when a limit is hit or the action succeeds. If someone could add a section for ACTIVITY that would be great as well, for people that don't want activity feedback.

Hornsoup: So I figured out how to turn it red, and then added a button in settings to make testing simpler and to fill out a player request, though I don't know where that request is anymore. PR is separate.

Hornsoup:
This should be good to go, I'll just keep working toward Death and update balance to just barely clear hurdles.

src/app/game-state/achievement.service.ts
```diff
@@ -443,19 +443,12 @@ export class AchievementService {
...
         return this.characterService.characterState.bloodlineRank >= 4;
       },
       effect: () => {
-        if (this.characterService.characterState.bloodlineRank == 4) {
-          this.characterService.characterState.bloodlineCost = 10000000000000
-        }
-        
-        if (this.characterService.characterState.bloodlineRank == 5) {
-          this.characterService.characterState.bloodlineCost = 1000000000000000
-        }
```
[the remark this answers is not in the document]
immortalityidle: Yeah, this was cleanup for existing save files at those ranks. If we're not changing the pricing this would be obsolete.

src/app/game-state/character.ts
```diff
@@ -387,13 +385,13 @@ export class Character {
...
+      const d = this.attributeScalingLimit + (this.attributeScalingLimit * 9 / 4) +
         (this.attributeScalingLimit * 90 / 20) +
-        (this.attributeSoftCap - (this.attributeScalingLimit * 100)) / 100;
-      const x = (aptitude - this.attributeSoftCap) * this.attributeScalingLimit / 10;
-
-      return (Math.pow(x, 0.1) * Math.log(x) + c) * empowermentFactor;
+        (this.attributeSoftCap - (this.attributeScalingLimit * 100)) / 100; // Pre-softcap
+        const x = (Math.pow((aptitude - this.attributeSoftCap) * Math.pow(this.attributeScalingLimit / 10000000000000, 0.15), 0.5) + d) * empowermentFactor; // Softcap
+        const c = 365 * 1000; // Hardcap
+      return (c / (- 1 - Math.log((x + c) / c))) + c; // soft-hardcap math
```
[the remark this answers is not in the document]
immortalityidle: Haha, every time this changes I understand it less and less.

src/app/game-state/character.ts
immortalityidle: Probably what I should have done in the first place. Yeah, not sure if it's useful now.

src/app/game-state/achievement.service.ts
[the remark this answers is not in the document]
Hornsoup: Since we can base the pricing purely on rank, it's possible to change them via the import now, so I did.

src/app/game-state/character.ts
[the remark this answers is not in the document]
Hornsoup: Yeah it's absolutely nuts in here. I literally just find a nifty curve and fudge the variable numbers until it's got a good balance, usually by testing and setting the goal roughly where it needs to be by checking it against the playground I have.

src/app/game-state/character.ts
Hornsoup: I mean technically you could use the old costs to derive the ranks, save the ranks, and then derive new costs from the ranks, but again, the costs and uses seem to be pretty set in.

immortalityidle (approved):

PR #90, opened 2022-07-24
Follower count

Hornsoup:
I spent hours trying to figure out how to make sorting simpler. Nothing worked as preferred, so I'm leaving this up to serve as a skeleton for someone with a brighter implementation.


Hornsoup:
I tried making the sorting block smaller with keys but I don't understand enough about them to work with the current Follower interface.

immortalityidle:
Good work. I stole a bunch of this and manually added it.

PR #91, opened 2022-07-25
Activity skipping + chains balance & fixes

Hornsoup:
This is broken. I don't know why, because sometimes it works, sometimes it doesn't, changing the skip loop and the new method in activity.service doesn't seem to fix anything.

And I forgot to do a separate branch for all this so here's a major change with a bunch of minor balance and fixes.

Ask away.

### DIFF
--- a/src/app/game-state/activity.service.ts
+++ b/src/app/game-state/activity.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { BattleService } from './battle.service';
 import { Activity, ActivityLoopEntry, ActivityType } from '../game-state/activity';
-import { AttributeType, CharacterAttribute } from '../game-state/character';
+import { CharacterAttribute, CharacterStatus } from '../game-state/character';
 import { CharacterService } from '../game-state/character.service';
 import { HomeService, HomeType } from '../game-state/home.service';
 import { InventoryService } from '../game-state/inventory.service';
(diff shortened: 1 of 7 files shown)


Hornsoup:
For example, the list will properly skip all actions if they're all set 0 repeat, but will perform a 0 action and skip a step of the next valid action, such that odd jobs 0 and Resting 2 will actually perform 1 odd job and 1 Resting. I have reviewed the code for awhile and I need fresh eyes on this.

src/app/game-state/activity.service.ts
```diff
@@ -689,27 +739,31 @@ export class ActivityService {
         if (numBuilders < 10){
           this.logService.addLogMessage("You fumble without the proper help and hurt yourself.","INJURY","EVENT");
           this.characterService.characterState.status.health.value -= this.characterService.characterState.status.health.max * 0.05;
+          this.mainLoopService.pause = true;
```
immortalityidle, line 742: Is this to help you debug? We for sure don't want to pause on every failure without some kind of checkbox to turn it off.

immortalityidle:
This is cutting deep into code that is very core to the game mechanics and also trying to add features and make misc. changes. Trying to do all that in the same pull request is making me very nervous. Can we separate out the resource utilization changes from everything else?

Hornsoup:
Definitely, it's more just here to get eyes on it since it's... well... not working even then. I'll remove everything not related to the skipping, brb.

src/app/game-state/activity.service.ts
Hornsoup: This is a deadly fail though. 20 days goes by and suddenly you wonder what happened. Though I do agree power should be in the players' hands in this, I've already messed with quite a bit as you've said.

src/app/game-state/activity.service.ts
```diff
@@ -551,8 +577,13 @@ export class ActivityService {
         } else {
           this.logService.addLogMessage("You fumble with the wrong tools and materials and hurt yourself.","INJURY","EVENT");
           this.characterService.characterState.status.health.value -= this.characterService.characterState.status.health.max * 0.05;
+          this.mainLoopService.pause = true;
         }
       }],
+      resourceUse: [{
+        stamina: 100,
+        health: this.characterService.characterState.status.health.max * 0.05 + 1
```
immortalityidle, line 585: I don't think this will do what I think you want it to. I think it will set it to 5% of the health value when the activity object is created, not the current health value when it gets checked.

PR #92, opened 2022-07-25
change time speed mechanics

1234abcdcba4321:
this is overall a nerf if you don't use banked time, but it feels way less arbitrary than what it was before. and now that you aren't living to 50k years anyway (probably) having time go way faster isn't as necessary.
(also, this is slabdrill)

### DIFF
--- a/src/app/game-state/main-loop.service.ts
+++ b/src/app/game-state/main-loop.service.ts
@@ -1,5 +1,7 @@
+
 import { Injectable,Injector } from '@angular/core';
 import { Subject } from 'rxjs';
+//import { threadId } from 'worker_threads';
 import { CharacterService } from './character.service';
 
 const TICK_INTERVAL_MS = 25;
@@ -9,6 +11,8 @@ export interface MainLoopProperties {
   unlockFastSpeed: boolean,
   unlockFasterSpeed: boolean,
   unlockFastestSpeed: boolean,
+  unlockAgeSpeed: boolean,
+  unlockPlaytimeSpeed: boolean,
   lastTime: number;
   tickDivider: number;
   pause: boolean;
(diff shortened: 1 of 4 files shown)


src/app/game-state/main-loop.service.ts
```diff
@@ -109,18 +115,39 @@ export class MainLoopService {
           }
           // and one extra for every 5000 years you've ever lived, up to 100 repeats
           repeatTimes += Math.min(Math.floor(this.totalTicks / 1825000), 100);
+        }*/
+
+        if (this.characterService && this.unlockAgeSpeed) {
+          ticksPassed *= Math.sqrt(1+this.characterService.characterState.age/73000);
```
immortalityidle, line 121: Comment for clarity please.

src/app/game-state/main-loop.service.ts
```diff
@@ -109,18 +115,39 @@ export class MainLoopService {
...
+        if (this.bankedTicks > 0 && this.useBankedTicks){
+          //using banked ticks makes time happen 10 times faster
+          ticksPassed *= 10;
+          this.bankedTicks -= timeDiff / TICK_INTERVAL_MS * 10;
+        }
+
+        this.tickCount += ticksPassed;
+        if (this.tickCount > 36500) {
+          //emergency lag prevention; this should never activate normally
+          this.tickCount = 36500;
```
immortalityidle, line 145: Should we re-bank the lost ticks in this case?

immortalityidle:
How significant of a nerf? And if you do use banked time how does it compare?

1234abcdcba4321:
If you use banked ticks it should exceed the old behavior in all cases except before you unlock the new manuals (though banked tick collection while not offline was slowed, so you might just not have as many banked ticks).
The nerf is pretty significant because the old behavior basically gave you a free 2x speed as soon as you were out of earlygame (unless banked ticks), which this definitely doesn't, and the formulas are just significantly harsher in general to make them smoother curves that don't need to have an explicit cap.

I don't think rebanking the lost ticks is that important; it's probably not going to be more than like 10 seconds worth of ticks, and if the game ever gets to the point where you're frequently hitting that failsafe then something's probably not right with game balance in general.

immortalityidle (approved):

Issue #57, opened 2022-07-20
Inventory Scrollbar moves down automatically


Ch1pless:
The inventory scrollbar moves a step down when the leftmost slot of any row occluded above by the inventory header is updated.

kagejohn:
I found a solution to this problem: https://stackoverflow.com/questions/48585358/how-do-i-prevent-chrome-from-scrolling-when-angular-5-rerenders-a-large-list

BeagleGaming1:
![image](https://user-images.githubusercontent.com/56142455/180668362-cef3781a-d06e-47b0-a087-a1e3432ced9b.png)
https://css-tricks.com/almanac/properties/o/overflow-anchor/

immortalityidle:
Fixed in 1.0.50.

PR #93, opened 2022-07-25
Inventory Scrollbar Moved On its own

kagejohn:
> Inventory Scrollbar moves down automatically #57 

Should now be fixed


immortalityidle (approved):
I love it when the fix is easy.

PR #94, opened 2022-07-25
A simpler option for checking resources

immortalityidle:
@Hornsoup here's what I was thinking as a simpler way to do the resource checking. Obviously incomplete and uses a ts-ignore hack to simplify the checking, but it works. Thoughts?

### DIFF
--- a/src/app/game-state/activity.service.ts
+++ b/src/app/game-state/activity.service.ts
@@ -108,7 +108,14 @@ export class ActivityService {
           }
         }
         const activity = this.getActivityByType(this.currentLoopEntry.activity);
-        activity.consequence[activity.level]();
+
+        // put some logic here checking if they have the unlock for this feature
+        if (this.checkResourceUse(activity)){
+          activity.consequence[activity.level]();
+        } else {
+          // put some logic here for when resting is unavailable
+          const activity = this.getActivityByType(ActivityType.Resting);
+        }
 
         // check for exhaustion
         if (this.characterService.characterState.status.stamina.value < 0) {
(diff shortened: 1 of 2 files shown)


src/app/game-state/activity.service.ts
```diff
@@ -108,7 +108,14 @@ export class ActivityService {
...
         }
         const activity = this.getActivityByType(this.currentLoopEntry.activity);
-        activity.consequence[activity.level]();
+
+        // put some logic here checking if they have the unlock for this feature
+        if (this.checkResourceUse(activity)){
+          activity.consequence[activity.level]();
+        } else {
+          // put some logic here for when resting is unavailable
+          const activity = this.getActivityByType(ActivityType.Resting);
```
immortalityidle, line 117: Oh, and actually do the consequence of the resting. Forgot to put that in.

Hornsoup: Hrm, if we're going this way, it'd be simple enough to just put this in between the consequence and activity, which would probably be let instead of const since it would be changed. 
if (!this.checkResourceUse(activity)){
          activity = this.getActivityByType(ActivityType.Resting);
}

Hornsoup: Not the point I'm trying to make... So it seems resting is replacing the current action, which means you'd want to have it as a kind of default action always available and for players to not bother adding it to the queue.

immortalityidle: You might still want to meditate just for the spirituality, and I do want this gated to pretty late in the game so we'd still want to maintain it as an activity to put in the queue, but once this is unlocked, sure, you could skip manually putting in rests and rely on this to do your resting as needed.

PR #96, opened 2022-07-26
Autorest & Impossible AutoPause & Sorting / Ascending & minor fixes

Hornsoup:
Oops, I probably put a bit too much in here while testing... Well, take a peek anyway.

Added autorest and related achievement, added Impossible Tasks autopausing, fixed the follower sorting wobble and added in asc/desc order, and some minor fixes for saving/loading and other such.

Probably so much due to testing and getting frustrated with small things that I fixed to make testing easier, as well as copying over some of the autorest stuff and finding that it included half the autopausing.

Everything is tested, with a minor bug in the autopausing that causes it to pause with negative health if you go step by step trying to kill yourself. Not really something a player would do normally.

### DIFF
--- a/src/app/follower-management-panel/follower-management-panel.component.html
+++ b/src/app/follower-management-panel/follower-management-panel.component.html
@@ -12,6 +12,9 @@
       <option [selected]="followerService.sortField === 'Power'">Power</option>
       <option [selected]="followerService.sortField === 'Remaining'">Remaining Life</option>
     </select>
+    <input type="checkbox" [checked]="this.Ascending" (change)="sortAscSwitch()">
+    <span *ngIf="this.followerService.sortAscending">Ascending</span>
+    <span *ngIf="!this.followerService.sortAscending">Descending</span>
   </span></div>
   <div class="scrollingDiv">
     <table>
--- a/src/app/follower-management-panel/follower-management-panel.component.html
+++ b/src/app/follower-management-panel/follower-management-panel.component.html
@@ -12,9 +12,8 @@
       <option [selected]="followerService.sortField === 'Power'">Power</option>
       <option [selected]="followerService.sortField === 'Remaining'">Remaining Life</option>
     </select>
-    <input type="checkbox" [checked]="this.Ascending" (change)="sortAscSwitch()">
-    <span *ngIf="this.followerService.sortAscending">Ascending</span>
-    <span *ngIf="!this.followerService.sortAscending">Descending</span>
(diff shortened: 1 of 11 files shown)


src/app/follower-management-panel/follower-management-panel.component.html
```diff
@@ -12,6 +12,9 @@ <h3>Manage Followers</h3>
       <option [selected]="followerService.sortField === 'Power'">Power</option>
       <option [selected]="followerService.sortField === 'Remaining'">Remaining Life</option>
     </select>
+    <input type="checkbox" [checked]="this.Ascending" (change)="sortAscSwitch()">
```
immortalityidle, line 15: Use followerService.sortAscending here instead of this.Ascending.

src/app/follower-management-panel/follower-management-panel.component.html
```diff
@@ -12,6 +12,9 @@ <h3>Manage Followers</h3>
...
+    <span *ngIf="this.followerService.sortAscending">Ascending</span>
```
immortalityidle, line 16: This seems like it would be confusing. If I uncheck it, I see Descending unchecked, which I would think would mean Ascending. I'd say just leave the label as Ascending.

src/app/follower-management-panel/follower-management-panel.component.ts
```diff
@@ -9,6 +9,7 @@ import { FollowersService } from '../game-state/followers.service';
 export class FollowerManagementPanelComponent implements OnInit {
 
   changeAll = 0;
+  Ascending = true;
```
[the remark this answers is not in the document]
Hornsoup: Oop, that's a leftover.

src/app/follower-management-panel/follower-management-panel.component.html
Hornsoup: Another leftover. Hrm. I guess I was on my last legs mentally doing this.

src/app/game-state/activity.service.ts
```diff
@@ -153,6 +174,28 @@ export class ActivityService {
 
   }
 
+  checkResourceUse(activity: Activity, spirit = false): boolean {
+    if (!activity.resourceUse || !activity.resourceUse[activity.level]){
+      return true;
+    }
+    if (spirit){
+      if (!activity.resourceUse[activity.level]["mana"]){
+        activity.resourceUse[activity.level]["mana"] = 0;
```
immortalityidle, line 183: I don't understand what this is doing here. Can you add a comment to clarify?

immortalityidle (approved):

PR #97, opened 2022-07-26
Impossible chains + minor fixes

Hornsoup:
Updated the task to be more in line with Determination and updated some saving and followermax conditions


src/app/game-state/activity.service.ts
```diff
@@ -237,7 +237,11 @@ export class ActivityService {
     this.currentApprenticeship = properties.currentApprenticeship || ActivityType.Resting;
     this.savedActivityLoop = properties.savedActivityLoop || [];
     this.autoRestUnlocked = properties.autoRestUnlocked || false;
-    this.pauseOnImpossibleFail = properties.pauseOnImpossibleFail || true;
+    if (properties.pauseOnImpossibleFail === undefined){
```
immortalityidle, line 240: Is there some situation when this would be undefined? Seems like we could handle this in the constructor and setProperties to ensure that its defined.

src/app/game-state/activity.service.ts
```diff
@@ -589,14 +593,22 @@ export class ActivityService {
       name: ['Forge Unbreakable Chain'],
       activityType: ActivityType.ForgeChains,
       description: ['Forge a chain strong enough to pull the island from the depths.'],
-      consequenceDescription: ['Uses 100 Stamina. If you have the right facilities and materials you might be able to create an unbreakable chain.'],
+      consequenceDescription: ['Uses 100 Stamina. If you have the right facilities, materials, and knowledge you might be able to create an unbreakable chain.'],
       consequence: [() => {
         this.characterService.characterState.status.stamina.value -= 100;
         const metalValue = this.inventoryService.consume('metal');
-        if (this.homeService.furniture.workbench && this.homeService.furniture.workbench.id === "anvil" && metalValue >= 150){
+        if (this.homeService.furniture.workbench && this.homeService.furniture.workbench.id === "anvil" && metalValue >= 150 && this.characterService.characterState.attributes.metalLore.value >= 1e9){
```
immortalityidle, line 600: Easier to just put the metalLore in the requirements?

immortalityidle: Oh. I see what you're doing, and I like it.

src/app/game-state/activity.service.ts
```diff
@@ -619,27 +631,27 @@ export class ActivityService {
       level: 0,
       name: ['Attach Chains to the Island'],
       activityType: ActivityType.AttachChains,
-      description: ['Swim deep and attach one of your chains to the island.'],
-      consequenceDescription: ['Uses 1000 Stamina. Requires an unbreakable chain.'],
+      description: ['Swim deep and attach one of your chains to the island, then pull.'],
+      consequenceDescription: ['Uses 1000000 Stamina when carrying something REALLY heavy to try hauling something really REALLY heavy. You\'re going to need an Unbreakable Chain and a long time to rest.'],
```
immortalityidle, line 635: Is that the stamina cap? If it is, maybe raise the cap.

src/app/game-state/activity.service.ts
```diff
@@ -619,27 +631,27 @@ export class ActivityService {
...
       consequence: [() => {
-        this.characterService.characterState.status.stamina.value -= 1000;
-        if (this.inventoryService.consume("chain") > 0){
-          this.logService.addLogMessage("You attach a chain to the island. and give your chains a tug.","STANDARD","EVENT");
+        if (this.inventoryService.consume("chain") > 0 ){
+          this.characterService.characterState.status.stamina.value -= 1000000;
```
immortalityidle, line 638: Does this let them succeed even if they don't have the stamina?

immortalityidle:
I'm not a huge fan of all-caps for emphasis, but it's not a deal-breaker if you're attached to it.

src/app/game-state/activity.service.ts
Hornsoup: I mean, is there some reason the cap needs raising? We literally have no use for huge huge stam other than me using it specifically for this.

src/app/game-state/activity.service.ts
Hornsoup: Crap, forgot to add that as a check since the old one didn't have that, definitely should. Had autorest on for testing like a derp.

Hornsoup: Hrm, now I'm thinking of making Pull Chains its own thing but meh, kinda feels extra.

Hornsoup:
Updated it with the Stamina check and fixed conflicts.

immortalityidle (approved):

PR #98, opened 2022-07-27
bugfixes and small changes

1234abcdcba4321:
added immortality ach, buffed infuse equipment (and have two unnecessary merge commits because i'm bad at git)


src/app/game-state/home.service.ts
```diff
@@ -767,7 +767,7 @@ export class HomeService {
       if (this.fields[i].daysToHarvest <= 0){
         let fieldYield = this.fields[i].yield;
         if (this.fields.length + this.extraFields > 300){
-          fieldYield = Math.floor((this.fields.length + this.extraFields) / 300);
+          fieldYield = Math.floor(fieldYield * (this.fields.length + this.extraFields) / 300);
```
immortalityidle, line 770: Is this right? This seems funny. @Hornsoup , second opinion please.

Hornsoup: Yeah, as is the fieldYield is completely overwritten, it should be multiplied by the factor of extrafields, so say you have 600 fields, right now this just sets the fieldYield to 2, instead of 2 * yield.

Hornsoup: Basically ignores all farming with 301+ fields.

immortalityidle (approved):

PR #103, opened 2022-07-27
cap time boost from lifespan

1234abcdcba4321:
also changed unlimited taels ach
(this cap was chosen because it's only immortal people who complain about the game running too fast)

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -454,11 +454,11 @@ export class AchievementService {
       unlocked: false
     },
     {
-      name: "Unlimited Taels",
-      description: "Your family has unlocked the secrets of compound interest. You probably never have to worry about money again.",
-      hint: "Family first. Especially in matters of money.",
+      name: "Not Unlimited Taels",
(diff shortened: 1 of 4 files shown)


src/app/game-state/achievement.service.ts
```diff
@@ -454,11 +454,11 @@ export class AchievementService {
       unlocked: false
     },
     {
-      name: "Unlimited Taels",
-      description: "Your family has unlocked the secrets of compound interest. You probably never have to worry about money again.",
-      hint: "Family first. Especially in matters of money.",
+      name: "Not Unlimited Taels",
```
immortalityidle, line 457: Is there a reason for changing this rather than just adding a new achievement? I'd rather just do both.

1234abcdcba4321: I don't see bloodline 4 as being any notable milestone that warrants being an achievement. There's no unlock tied to it, and it's not like the fact that it's infinite money is something particularly hard to figure out.

immortalityidle (approved):

PR #106, opened 2022-07-28
Added Lore Balance Activity

Hornsoup:
Also reordered activities so they were a little more relative to each other and lore grouped.

### DIFF
--- a/src/app/game-state/activity.service.ts
+++ b/src/app/game-state/activity.service.ts
@@ -486,29 +486,32 @@ export class ActivityService {
       newList.push(this.MoveStars);
     }
 
-    newList.push(this.OddJobs);
     newList.push(this.Resting);
+    newList.push(this.OddJobs);
     newList.push(this.Begging);
+    newList.push(this.Burning);
     newList.push(this.Farming);
-    newList.push(this.GatherHerbs);
-    newList.push(this.ChopWood);
     newList.push(this.Mining);
     newList.push(this.Smelting);
     newList.push(this.Blacksmithing);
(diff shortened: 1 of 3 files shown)


src/app/game-state/character.ts
```diff
@@ -426,6 +426,16 @@ export class Character {
     return increaseAmount;
   }
 
+  decreaseAttribute(attribute: AttributeType, amount: number): number {
```
immortalityidle, line 429: This didn't end up being used, right?

src/app/game-state/activity.service.ts
```diff
@@ -486,29 +486,32 @@ export class ActivityService {
...
+    newList.push(this.Alchemy);
     newList.push(this.BodyCultivation);
     newList.push(this.MindCultivation);
-    newList.push(this.CoreCultivation);
-    newList.push(this.InfuseEquipment);
-    newList.push(this.InfuseBody);
-    newList.push(this.ExtendLife);
+    if(this.characterService.characterState.manaUnlocked || this.characterService.characterState.easyMode){
+      newList.push(this.ManaControl);
+      newList.push(this.CoreCultivation);
```
immortalityidle, line 508: Can we just put ManaControl in the if?

src/app/game-state/character.ts
Hornsoup: Ah, forgot about that method.

Hornsoup:
Took care of review points.

immortalityidle (approved):

PR #107, opened 2022-07-29
fix breaks are bad ach

1234abcdcba4321:
did some trivial stuff and also made the associated manual a bit better


src/app/game-state/activity.service.ts
```diff
@@ -133,6 +133,10 @@ export class ActivityService {
         const rest = this.getActivityByType(ActivityType.Resting);
         if (!this.checkResourceUse(activity) && rest.unlocked && this.autoRestUnlocked){ // check for resources, rest activity is available, and autopause unlocked.
           activity = rest;
+          if (this.checkMaxResourceUse(activity)) {
+            //don't skip the current activity unless you don't have enough max resources to fit it
```
immortalityidle, line 137: Hmm. Not sure if I like this better. This would stall you at this activity until you've rested enough to do it, right? Thought @Hornsoup ?

Hornsoup: This would actually do nothing, since the activity would change to rest and then have it checked for the 0 requirements for rest. None of this code is useful.

src/app/game-state/activity.service.ts
```diff
@@ -154,7 +158,6 @@ export class ActivityService {
           this.characterService.characterState.status.health.value -= 0.01 * this.characterService.characterState.status.health.max;
         }
         // check for activity death
-        this.activityDeath = false;
```
Hornsoup, line 157: This is a required false reduction. Removing it means every death thereafter would be considered an activity death. It's a single activity switch. This has to be included.

src/app/ascension-store-modal/ascension-store-modal.component.html
```diff
@@ -55,6 +55,6 @@
     </tr>
   </table>
   <span *ngIf="characterService.characterState.empowermentFactor > 1">
-    You have consumed {{(characterService.characterState.empowermentFactor - 1) * 100 | number:'1.0-0'}} Empowerment Pills which multiplies all your attribute gains by {{characterService.characterState.getEmpowermentMult()  | number:'1.0-3'}}. You will always keep this bonus, even after ascending.
+    You have consumed {{(characterService.characterState.empowermentFactor - 1) * 100 | number:'1.0-0'}} Empowerment Pills which multiplies all your attribute gains by {{characterService.characterState.getEmpowermentMult()  | number:'1.0-3'}} (except you cannot surpass human limits). You will always keep this bonus, even after ascending.
```
Hornsoup, line 58: Probably something more like *subject to human limitations

src/app/game-state/activity.service.ts
Hornsoup: It's effectively the else statement of a true false if switch.

src/app/game-state/activity.service.ts
Hornsoup: Actually, yeah, with the later currentTickCount reduction that would just rest without passing days IIUC?

src/app/game-state/activity.service.ts
1234abcdcba4321: Ah, whoops, didn't notice the variable being used somewhere other than for the achievement unlock.

src/app/ascension-store-modal/ascension-store-modal.component.html
1234abcdcba4321: I don't think that's clear enough; 100000 times 10 is still 1000000 unless it's clear that the number given is a lie. The usage of "except" is supposed to help clarify this.

1234abcdcba4321:
Okay, so adding the odd jobs thing in this PR was accidental but I'm not sure how to get rid of it, so it's there now I guess.
Anyway, fixed the issues that got pointed out.

src/app/game-state/activity.service.ts
immortalityidle: I don't think this messes with passing days, though that was my first instinct when I looked at it. Then I remembered that variable just counts how far along in the repeats for that activityLoop entry we are. Days still pass, we just don't advance to the next activity until we've rested enough to do this one (unless we'll never be able to, in which case it skips with the maxResource check).

immortalityidle:
Hmm. Now I have to think about the odd jobs updates. What's your thinking on what their role is?

src/app/game-state/activity.service.ts
Hornsoup: It's still not reset to false for this, since if it's stuck true then future battle deaths etc will give the wrong death message.

src/app/game-state/activity.service.ts
Hornsoup: Ok, I think I understand what's tricking me. It's the second maxresourceuse check after the rest. It will always reduce the currenttickcount, causing it to potentially get stuck at 0.

src/app/game-state/activity.service.ts
```diff
@@ -1207,11 +1213,111 @@ export class ActivityService {
...
+          if (this.characterService.characterState.attributes[attribute].value < this.characterService.characterState.attributes[lowStat].value) {
+            lowStat = attribute;
+          }
+          if (this.characterService.characterState.attributes[attribute].value > this.characterService.characterState.attributes[highStat].value) {
+            highStat = attribute;
+          }
+        }
+        let statSum = 0;
+        for (const attribute of ["strength","toughness","speed","intelligence","charisma"] as AttributeType[]){
+          if (attribute === lowStat) this.characterService.characterState.increaseAttribute(attribute, 300);
```
Hornsoup, line 1284: I don't think you're aware, but this is multiplied by the aptitude multiplier. These numbers are usually 1 or less... @immortalityidle I feel like you could potentially manually enter parts of this PR but as it is there's a lot of needless additions and there's the comment about a lack of testing.

immortalityidle:
This PR does a lot and makes me nervous. Can you please break it into multiple PRs, one per significant feature change/bugfix?

PR #110, opened 2022-07-30
Easy Mode Confirmation

Hornsoup:


src/app/options-modal/options-modal.component.ts
```diff
@@ -117,9 +117,13 @@ export class OptionsModalComponent {
   }
 
   easyModeChange(event: Event){
-    if (!(event.target instanceof HTMLInputElement)) return;
-    this.gameStateService.easyModeEver = true;
-    this.characterService.characterState.easyMode = event.target.checked;
+    event.preventDefault();
+    if (!this.characterService.characterState.easyMode && confirm("This will enable easy mode and mark your save permanently. Are you sure?")){
```
immortalityidle, line 121: I'd use the easyModeEver flag for this check. If they're turning if back on after turning it on and off it's not a big deal.

src/app/options-modal/options-modal.component.ts
```diff
@@ -117,9 +117,13 @@ export class OptionsModalComponent {
...
+      this.gameStateService.easyModeEver = true;
+      this.characterService.characterState.easyMode = !this.characterService.characterState.easyMode;
+    } else if (confirm("Turning off easy mode won't take off the mark on your save. Are you sure? Wouldn't you rather Hard Reset?")){
```
immortalityidle, line 124: I don't think this needs to be a confirm. An alert is fine, or just putting it in as some text in an ngIf controlled span that only shows when easy mode is on. Maybe change text to something along the lines of "Only starting your journey to immortality from the beginning again (Hard Reset) will prove you never took the easy path."

Hornsoup:
Added requests, all of them seemed pretty no duh when actually reviewing.

immortalityidle (approved):

PR #112, opened 2022-07-31
Ore/Wood Update

Hornsoup:
Includes:
- Woodworking / Blacksmithing / Leatherworking update to roughly coincide with names for weapons and armor
- Follower update to coincide with these changes
- Update to impact of Strength and Toughness on weaponry, minor multiplier of log10, but players seeing numbers going up is the point.
- Update to accuracy formula to incite a need for Speed stat (about 400 Billion by 9/9/3, doable in testing)
- optional pause on thug, visible/clickable with home settings (thus requires a manual)
- Minor update to recruiting, because only an idiot would use up the last of their money on advertising guaranteed to have no effect.


Hornsoup:
Figured out how to reduce that armorer block. Had to use non-null assertions since the compiler was being a butt and not recognizing the logic, that's all.

src/app/game-state/battle.service.ts
```diff
@@ -208,8 +208,8 @@ export class BattleService {
   }
 
   youAttack(){
+    this.characterService.characterState.accuracy = Math.min((this.troubleKills + Math.sqrt(this.characterService.characterState.attributes.speed.value)) / this.troubleKills / 2, 1)
```
immortalityidle, line 211: Is this going to fight back and forth with the accuracy setting in the recalculateDerivedStats function?

immortalityidle: Oh. Nevermind. You removed that. Any reason we set this here and not there?

src/app/game-state/item-repo.service.ts
```diff
@@ -625,39 +625,53 @@ export class ItemRepoService {
       value: 9,
       description: 'A spiritual quality log.',
     },
+    redwoodLog: {
```
immortalityidle, line 628: Redwood is pretty soft. Seems like this should be a lower cost wood. Let's give these unique descriptions too.

src/app/game-state/battle.service.ts
Hornsoup: Because it's easier to use troubleKills from here rather than having to inject battle into character.

src/app/game-state/item-repo.service.ts
Hornsoup: Oop, forgot descriptions. Not sure where to place the woods, because some are useful for structures and others have a way different janka scale than what's already here.

Hornsoup: Kinda want to make elm balsa instead.

Hornsoup: Alright, roughed some locations based on softness since harder woods are more suitable for weapons, existing woods were really wherever they wanted to be.

src/app/game-state/item-repo.service.ts
```diff
@@ -562,102 +562,116 @@ export class ItemRepoService {
...
-    },
-    blackwoodLog: {
-      id: 'blackwoodLog',
-      name: 'blackwood log',
-      type: 'wood',
-      value: 8,
+      value: 9,
       description: 'An amazing quality log.',
     },
-    peachwoodLog: {
```
immortalityidle, line 621: No objections to the new wood, but can we add peachwood back in there too? That's an actual historical and mythological thing that I liked.

src/app/game-state/item-repo.service.ts
```diff
@@ -562,102 +562,116 @@ export class ItemRepoService {
       value: 1,
       description: 'A basic animal hide.'
     },
-    elmLog: {
```
immortalityidle, line 565: Can we keep both?

src/app/game-state/activity.service.ts
```diff
@@ -1783,7 +1783,7 @@ export class ActivityService {
               const grade = this.inventoryService.consume('wood');
               if (grade >= 1){ // if the wood was found
                 this.inventoryService.addItem(this.inventoryService.generateWeapon(
-                  (grade * 5) + Math.floor(Math.log2(this.characterService.characterState.attributes.woodLore.value)), 'wood'));
+                  Math.floor(Math.pow(this.characterService.characterState.attributes.metalLore.value, grade/32)), 'wood'));
```
kagejohn, line 1786: how come it was changed to metal lore for the wood weapon?

Hornsoup: crapity crap that's copypasta

src/app/game-state/item-repo.service.ts
Hornsoup: I'll drop out redwood for balsa and rubywood for lignum vitae, maybe that's best.

Hornsoup:
That should be all the woodworking fixes, anything else I missed?

src/app/game-state/item-repo.service.ts
immortalityidle: Sure. Not opposed to adding more wood types if you want though.

Hornsoup:
Crap, just realized how much impact this would have on current weapons being OP and newer players getting hit hard, and how much impact equipment infusion has.

Hornsoup:
Updated a lot of the weapon and armor related balance and moved the gem usage into the weapon and armor generation since it was basically a huge method and added activity that weren't necessary since they did the same thing that automatically using gems was for.

immortalityidle (approved):

PR #118, opened 2022-08-01
MACOS: Use ctrl OR cmd click due to OS reservation

jmrd98:
Very simple adjustment here. I use a macbook to play the game, but ctrl-click gets reserved at the OS level to perform a right click action. (Or vice versa depending on perspective)

This PR just adds an OR event.metaKey to the places where event.ctrlKey is used. I didn't make any interface changes, because I didn't know how that might want to be handled up stream, but at least the functionality is accessible now.


immortalityidle (approved):

immortalityidle:
Thanks. We'll get it into the next build. Are you on the discord so I can give you dev credit?

jmrd98:
Ah, awesome. 

I mean, technically? I just hopped on yesterday to see if there was an existing fix. I’d be under Komodo #7176

PR #123, opened 2022-08-02
ETA for buildings

Hornsoup:


src/app/home-panel/home-panel.component.html
```diff
@@ -31,9 +31,10 @@ <h3 class="panelHeader">
               tooltip="Buy Furniture" (click)="storeClicked()">
               shopping_cart
             </mat-icon>
-            <div *ngIf="homeService.upgrading" class="progress-bar" tooltip="Upgrading your home.">
-              <span [style.width.%]="100 * homeService.houseBuildingProgress"></span>
+            <div *ngIf="homeService.upgrading" class="progress-bar" style="height:20px;" tooltip="Upgrading your home." [attr.value]=buildTimeYears()>
+              <span [style.width.%]="195 * homeService.houseBuildingProgress"></span>
```
immortalityidle, line 35: I don't understand the 195.

immortalityidle: OK. I see 195 is the width, but isn't this supposed to be a percentage?

Hornsoup: That would be because I understand very little about frontend lol... changed to 100.

immortalityidle (approved):

PR #124, opened 2022-08-02
Performance fixes

Hornsoup:
- Removed deep copy of enemies via JSON as it was unnecessary. From what I can tell the reference is passed into the array and the garbage collection doesn't take it when the method returns due to the higher scope of the array, and thus does not impact gameplay.
- Moved description updaters to subscribers as they weren't necessary every tick.
- Moved Math to a frame subscriber to avoid excess tiny increment math calls from high tick counts.

### DIFF
--- a/src/app/game-state/battle.service.ts
+++ b/src/app/game-state/battle.service.ts
@@ -287,7 +287,7 @@ export class BattleService {
     this.currentEnemy = enemyStack;
   }
 
-  addEnemy(enemy: Enemy){
+  addEnemy(enemy: Enemy, repoEnemy = true){
     this.logService.addLogMessage("A new enemy comes along to trouble your sleep: " + enemy.name, 'STANDARD', 'COMBAT');
     for (const enemyIterator of this.enemies) {
       if (enemyIterator.enemy.name === enemy.name) {
@@ -299,7 +299,11 @@ export class BattleService {
       }
     }
     // it didn't match any, create a new enemyStack
-    this.enemies.push({enemy: enemy, quantity: 1});
+    if(repoEnemy){
+      this.enemies.push({enemy: JSON.parse(JSON.stringify(enemy)), quantity: 1});
+    } else {
+      this.enemies.push({enemy: enemy, quantity: 1});
+    }
     if (this.currentEnemy === null){
       this.currentEnemy = this.enemies[0];
     }
@@ -335,7 +339,7 @@ export class BattleService {
       attack: this.troubleKills / 5,
       defense: this.troubleKills / 5,
       loot: [gem]
-    });
+    }, false);
     this.troubleKills++;
   }
 
--- a/src/app/game-state/battle.service.ts
+++ b/src/app/game-state/battle.service.ts
@@ -287,7 +287,7 @@ export class BattleService {
     this.currentEnemy = enemyStack;
   }
 
-  addEnemy(enemy: Enemy, repoEnemy = true){
+  addEnemy(enemy: Enemy, repoEnemy = false){
     this.logService.addLogMessage("A new enemy comes along to trouble your sleep: " + enemy.name, 'STANDARD', 'COMBAT');
     for (const enemyIterator of this.enemies) {
       if (enemyIterator.enemy.name === enemy.name) {
@@ -339,7 +339,7 @@ export class BattleService {
       attack: this.troubleKills / 5,
       defense: this.troubleKills / 5,
       loot: [gem]
-    }, false);
+    });
     this.troubleKills++;
   }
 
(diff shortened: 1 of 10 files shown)


src/app/game-state/battle.service.ts
```diff
@@ -301,7 +299,7 @@ export class BattleService {
       }
     }
     // it didn't match any, create a new enemyStack
-    this.enemies.push({enemy: JSON.parse(JSON.stringify(enemy)), quantity: 1});
+    this.enemies.push({enemy: enemy, quantity: 1});
```
immortalityidle, line 302: The reason I did a deep copy was so that I could damage the enemy without it damaging the enemy template for things like wolves and armies. Is that no longer an issue?

src/app/game-state/main-loop.service.ts
```diff
@@ -98,6 +99,10 @@ export class MainLoopService {
       this.longTickSubject.next(true);
     }, LONG_TICK_INTERVAL_MS);
 
+    window.setInterval(()=> {
+      this.frameSubject.next(true);
```
immortalityidle, line 103: I don't understand the purpose of doing an additional interval here on the same frequency as the standard tick.

src/app/game-state/character.ts
```diff
@@ -79,9 +80,19 @@ const INITIAL_AGE = 18 * 365;
...
 
-  constructor(private logService: LogService,
-    private titlecasePipe: TitleCasePipe){
-  }
+  constructor(
+    private logService: LogService,
+    private titlecasePipe: TitleCasePipe,
+    public mainLoopService: MainLoopService
+    ){
+      mainLoopService.frameSubject.subscribe(() => {
```
immortalityidle, line 88: Seems like this could be a regular tick subscription and have the same effect since this new tick is on the same interval period.

src/app/game-state/character.ts
```diff
@@ -114,83 +126,95 @@ export class Character {
       value: 1,
       lifeStartValue: 1,
       aptitude: 1,
+      aptitudeMult: 1,
```
immortalityidle, line 129: Legacy save files won't have these, so we may need something in the setProperties to make sure they exist before they get used.

src/app/game-state/character.ts
Hornsoup: It's not though. The tick subscriber calls Tick, which passes the Subject.next repeatedly for every tick. This only broadcasts it once per frame. I was able to get the 100 years of a particularly bogged save down to 1 second from 1.4 with this change alone so I know there's an effect.

src/app/game-state/main-loop.service.ts
Hornsoup: Because the standard tick calls Tick and thereby tickSubject.next on a loop.

src/app/game-state/battle.service.ts
Hornsoup: AH, that helps with my understanding, I didn't think about the exact ramifications on the repo. I only tried one stack of mice during testing. Maybe I can add a parameter to keep deep copies to the repo so that the generated enemies can have shallow copies and reduce the performance loss. That should work.

src/app/game-state/character.ts
Hornsoup: Right, I'll probably turn it into a method since it's really just a derived stat.

Hornsoup:
Also made slight changes to hell per the addEnemy use there, and of course ESLint felt the need to automatically make lets into consts

immortalityidle:
Sorry, conflicted it with new hell stuff. Can you resolve?

Hornsoup:
Okay, added the false flags for those new enemies while resolving as well.

Hornsoup:
Hrm... Maybe it should be flipped so it defaults false, and later I can remove the enemyrepo entirely? Do you want to keep it?

immortalityidle:
> Hrm... Maybe it should be flipped so it defaults false, and later I can remove the enemyrepo entirely? Do you want to keep it considering the performance issues with deep-copying from it?

I'm not married to it. When I set it up I thought I'd use it a lot more.

Hornsoup:
Alright, I'll take care of that with the later performance fixes I'll work on after work.

src/app/game-state/main-loop.service.ts
immortalityidle: I think this is much faster than angular's interpolation rate, but I'm having trouble finding anything that says how often interpolation actually runs. Crazy idea: tie the frameSubject.next() to a function that gets called by the interpolator so that it actually fires on angular interpolation frames instead of on a fixed interval.

immortalityidle (approved):

PR #125, opened 2022-08-02
Optimising followers

kagejohn:
I found that these 3 follower types caused lag: farmer, hunter, scout.
I have fixed farmer and hunter.
Scout is fixed in #124 

At the same time i found a bug with field.yield.


immortalityidle:
This is much better than what I put in.

kagejohn:
That should be all of the places where a follower is updated in some way

immortalityidle (approved):

immortalityidle:
Oh, one more should probably call the set total power function at the bottom of setProperties.

kagejohn:
Yeah, i noticed that followers didn't do anything on page load, so i panicked.
Thanks for the location.

PR #130, opened 2022-08-03
Update for inactive and removing limiters

Hornsoup:
- Added a check for bankedTicks overuse, which also requires the removal of the now defunct lag prevention (since the loop exits after a certain amount of time) to function correctly.

### DIFF
--- a/src/app/game-state/main-loop.service.ts
+++ b/src/app/game-state/main-loop.service.ts
@@ -118,18 +118,22 @@ export class MainLoopService {
       } else {
         const currentTPS = this.getTPS(this.tickDivider) / 1000 * TICK_INTERVAL_MS;
         const topTPS = this.getTPS(this.topDivider) / 1000 * TICK_INTERVAL_MS;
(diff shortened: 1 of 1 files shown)


src/app/game-state/main-loop.service.ts
```diff
@@ -143,11 +147,7 @@ export class MainLoopService {
           }
           this.tickCount = 0;
         }
-        // if the game is inactive for longer periods of time the banked ticks might go negative.
-        if (this.bankedTicks < 0) {
-          this.bankedTicks = 0;
-          this.useBankedTicks = false;
-        }
+        this.tickCount += inactiveTickHolder;
```
immortalityidle, line 150: Any risk of this picking up a huge number that would bog it down?

Hornsoup: It simply can't due to the time limiter on the loop, if it goes over it just adds it to the bankedticks, actually tested and got it to do that.

PR #132, opened 2022-08-04
Performance: Kicked formatNumber out of BigNumberPipe

Hornsoup:


immortalityidle:
This a big performance gain? Or do we just not like formatNumber?

Hornsoup:
It definitely improved performance, from about 20k days per second to about 26k days when I was testing full throttle.

immortalityidle (approved):

PR #133, opened 2022-08-04
Improve LogService performance

Hornsoup:
Tore most of it out, works the same as before, just less looping and unshifts. Review to follow

### DIFF
--- a/src/app/game-state/log.service.ts
+++ b/src/app/game-state/log.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
 import { isEmpty } from 'rxjs';
+import { MainLoopService } from './main-loop.service';
 
 const LOG_MERGE_INTERVAL_MS = 1000;
 export type LogType = 'STANDARD' | 'INJURY';
@@ -35,7 +36,13 @@ export class LogService {
   followerLog: Log[] = [];
   currentLog: Log[] = [];
 
-  constructor() {
+  constructor(
+    
+    mainLoopService: MainLoopService
+  ) {
+    mainLoopService.frameSubject.subscribe(() => {
+      this.updateLogTopics();
+    });
     this.addLogMessage("Once in a very long while, a soul emerges from the chaos that is destined for immortality. You are such a soul.", 'STANDARD', 'STORY');
     this.addLogMessage("Your journey to immortality begins as a humble youth leaves home to experience the world. Choose the activities that will help you cultivate the attributes of an immortal.", 'STANDARD', 'STORY');
     this.addLogMessage("It may take you many reincarnations before you achieve your goals, but with each new life you will rise with greater aptitudes that allow you to learn and grow faster.", 'STANDARD', 'STORY');
@@ -63,10 +70,7 @@ export class LogService {
 
     if (log.length === 0 || ((newMessage.timestamp - log[0].timestamp) > LOG_MERGE_INTERVAL_MS) || !log[0].message.includes(newMessage.message)) {
       // Initialization || Repeat Not Found || Repeat is not within 1 second
-      log.unshift(newMessage);
-      if (this.logTopics.includes(topic)) {
-        this.addToCurrentLog(newMessage);
-      }
+      log.push(newMessage);
     } else {
       // Repeat Found
       const hasRepeatNumber = /\((\d+)\)$/.exec(log[0].message);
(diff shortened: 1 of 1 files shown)


src/app/game-state/log.service.ts
```diff
@@ -135,53 +127,38 @@ export class LogService {
...
-        const timestamp = loopLog[loopLog.length - 1].timestamp || 0;
-        if (timestamp < latestTimestamp){
-          latestTimestamp = timestamp;
-          latestLog = loopLog;
-        }
-      }
-      this.addToCurrentLog(latestLog[latestLog.length - 1]);
-      latestLog.splice(latestLog.length - 1, 1);
-    }
+    logs.sort((a,b) => b.timestamp - a.timestamp);
```
Hornsoup, line 160: Why do all this looping when you can just sort?

src/app/game-state/log.service.ts
```diff
@@ -135,53 +127,38 @@ export class LogService {
...
+    this.currentLog = logs.slice(0,299);
```
Hornsoup, line 161: Instead of emptying currentLog, just give it the most recent slice and let the garbage collection sweep the old one.

src/app/game-state/log.service.ts
```diff
@@ -35,7 +36,13 @@ export class LogService {
...
   currentLog: Log[] = [];
 
-  constructor() {
+  constructor(
+    
+    mainLoopService: MainLoopService
+  ) {
+    mainLoopService.frameSubject.subscribe(() => {
+      this.updateLogTopics();
+    });
```
Hornsoup, line 45: Update along with the tick limiter. Tried using {{method()}} but that threw a ton of errors.

src/app/game-state/log.service.ts
```diff
@@ -63,10 +70,7 @@ export class LogService {
 
     if (log.length === 0 || ((newMessage.timestamp - log[0].timestamp) > LOG_MERGE_INTERVAL_MS) || !log[0].message.includes(newMessage.message)) {
       // Initialization || Repeat Not Found || Repeat is not within 1 second
-      log.unshift(newMessage);
-      if (this.logTopics.includes(topic)) {
-        this.addToCurrentLog(newMessage);
-      }
+      log.push(newMessage);
```
Hornsoup, line 73: Ripped out addtoCurrentLog since it was pushing one at a time and just did a full sort and slice.

src/app/game-state/log.service.ts
```diff
@@ -80,9 +84,6 @@ export class LogService {
     }
 
     // check if we need to age off the oldest logs
-    if (log.length > 100 && topic !== 'STORY'){
-      log.splice(100, 1);
-    }
```
Hornsoup, line 85: No more single splicing, do it every frame as a group.

src/app/game-state/log.service.ts
```diff
@@ -135,53 +127,38 @@ export class LogService {
...
       this.newCrafting = "";
-      logs.push([...this.craftingLog]);
+      this.craftingLog = this.craftingLog.slice(-300)
+      logs.push(...this.craftingLog);
     }
     if (this.logTopics.includes('FOLLOWER')){
       this.newFollower = "";
-      logs.push([...this.followerLog]);
+      this.followerLog = this.followerLog.slice(-300)
+      logs.push(...this.followerLog);
```
Hornsoup, line 154: Moved slicing down into these.

src/app/game-state/log.service.ts
```diff
@@ -98,15 +99,6 @@ export class LogService {
...
 
-  /** Add Message To Current Log */
-  private addToCurrentLog(newMessage: Log): void {
-    // Maximum Log Length of 300
-    if (this.currentLog.length >= 300){
-      this.currentLog.pop();
-    }
-    this.currentLog.unshift(newMessage);
-  }
-
```
Hornsoup, line 109: Don't need it, updateLogTopics() just does a sort and slice.

src/app/game-state/log.service.ts
```diff
@@ -35,7 +36,13 @@ export class LogService {
   followerLog: Log[] = [];
   currentLog: Log[] = [];
 
-  constructor() {
+  constructor(
+    
+    mainLoopService: MainLoopService
+  ) {
+    mainLoopService.frameSubject.subscribe(() => {
+      this.updateLogTopics();
```
immortalityidle, line 44: This seems like it does a lot of extra sorting. How's the performance compare? And could we optimize further by adding a flag that gets set in addLogMessage so we know if there's new stuff to sort or if we can skip it this tick?

Hornsoup: This isn't a tick, it's a 25ms, so hundreds of ticks could go by. Let's not forget that when the tickSubject triggers it loops its next method. And yeah, this improves a lot, about as much as the BigNumber pipe fix. I'd love to be able to do this update once per frame, preferably pre-draw, but I don't know what to listen to. As is, it's roughly twice a frame.

PR #134, opened 2022-08-04
Fix for BigNumberPipe

kagejohn:
It is possible that i am wrong but the other version added and removed zeros


Hornsoup:
You're right, I was testing with a full set and didn't think about what numberPower represents in its raw form so I didn't notice it.

immortalityidle (approved):

PR #138, opened 2022-08-05
Added inventory sort button.

Hornsoup:
Sort priority is type, value, then quantity, then name. Kinda want to subgroup the materials and toss gems to the bottom but there's no real way to easily do so based on item properties.

### DIFF
--- a/src/app/game-state/inventory.service.ts
+++ b/src/app/game-state/inventory.service.ts
@@ -315,6 +315,26 @@ export class InventoryService {
     }
   }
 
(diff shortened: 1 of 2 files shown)


immortalityidle:
Shouldn't be too bad to add sorting categories to items as an optional property if you want. A little tedious, but not complicated.

Hornsoup:
I just tossed the gems and equips into their own temp arrays, should make for less wobbly inventory.

immortalityidle (approved):

PR #141, opened 2022-08-07
Added toggles for autoselling various old items

Hornsoup:


Hornsoup:
Oop, accidentally deleted...

PR #146, opened 2022-08-08
Added an option to configure autosave interval

OneMoreGuy:
From my tests, the option is retained even after reloading. I'm not the best coder though, so please check if it works.


immortalityidle:
Wow. This is a massive rewrite for what should be a very small change. Why are we making a whole new service for this instead of just making it a single property of GameStateService?

PR #148, opened 2022-08-09
Multiple Save Files

Hornsoup:
Here's the work so far, though there's some bugs to work out.


immortalityidle:
I still think this is more complicated than it needs to be, but I don't think it's going to hurt anything.

Hornsoup:
I totally think so too, honestly. I'll probably cut back the naming and just make numbers, I mostly wanted to experiment with how it works. There's some funky bugs in here with this implementation though.

PR #150, opened 2022-08-09
Configurable autosave interval

OneMoreGuy:
Added the option to configure the autosave interval from 1 to 900 seconds (15 min). Change persisted on refresh and on completely closing the cmd instance + chrome window.


OneMoreGuy:
Only the last commit is the actual change, the rest are me cleaning up remnants of failed attempts.

immortalityidle (approved):

PR #152, opened 2022-08-09
File import and export

kagejohn:
https://discord.com/channels/996414713766363146/998578984730251276/1006513197483831296


immortalityidle:
Does this get rid of the need for the import/export panel?

kagejohn:
It doesn't, it just adds 2 more options.
We could remove the input box so that it is no longer possible to get it as a string.

immortalityidle (approved):

PR #159, opened 2022-08-12
Fix: Not updating descriptions after Infusion

Hornsoup:
Now updates after infusing equipment without autoequipbest enabled.


Hornsoup:
- Now reloads to update the scientific numbers. Pipes are lazy bad mkay?
- Fixed isImport flag not updating to false.

immortalityidle (approved):

PR #163, opened 2022-08-17
Potential slow speed issue with banked

Hornsoup:
In case someone suspends on slow with banked checked I suppose.


Hornsoup:
Still trying to see what part of the code is screwy with offline ticks and suspending.

immortalityidle (approved):

PR #172, opened 2022-08-27
Missing if for spiritActivity found

kagejohn:
I am fairly sure there is supposed to be an if there.


immortalityidle (approved):

immortalityidle:
Good catch, thanks!

PR #174, opened 2022-11-20
Fix "Hoarder" achievment

vhaldemar:
Right now "Hoarder" unlocks almost simultaneously with "Landfill", it is probably a bug


immortalityidle (approved):
Not what I was originally thinking, but I do like spacing out the achievements. Thanks. I'll get it into the next version update.

PR #176, opened 2023-02-05
Logging refactor

LiquidAlmond:
Refactored logging code to allow for easily adding or modifying log topics and put this logging more inline with existing libraries.

The last bit could just be my opinion though.

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -1,5 +1,5 @@
 import { Injectable, Injector } from '@angular/core';
-import { LogService } from './log.service';
+import { LogService, LogTopic } from './log.service';
 import { CharacterService } from './character.service';
 import { InventoryService } from './inventory.service';
 import { HomeService, HomeType } from './home.service';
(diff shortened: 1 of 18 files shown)


src/app/game-state/character.ts
```diff
@@ -573,7 +573,7 @@ export class Character {
       increaseAmount = amount;
     }
     this.attributes[attribute].value += increaseAmount;
-    if (!this.highestAttributes[attribute] || this.highestAttributes[attribute] > this.attributes[attribute].value) {
+    if (!this.highestAttributes[attribute] || this.highestAttributes[attribute] < this.attributes[attribute].value) {
```
LiquidAlmond, line 576: Messed up the commits on this. Change in another PR that should be merged before this one.

src/app/game-state/achievement.service.ts
```diff
@@ -653,7 +653,7 @@ export class AchievementService {
       description: "You have established an empire that will never fall, and a bloodline that will always inherit it.",
       hint: "Bloodline Empire.",
       check: () => {
-        return (this.homeService.home.type === HomeType.Capital && this.characterService.characterState.bloodlineRank >= 7);
+        return (this.homeService.home.type >= HomeType.Capital && this.characterService.characterState.bloodlineRank >= 7);
```
LiquidAlmond, line 656: Messed up the commits on this. Change in another PR that should be merged before this one.

immortalityidle (commented):
I like these changes generally, but how does the performance compare to the existing code? One of the late-game issues we've seen before was with logs bogging down the game as time accelerated, so I want to make sure that any changes are at least as performant as what we have now.

LiquidAlmond:
> I like these changes generally, but how does the performance compare to the existing code? One of the late-game issues we've seen before was with logs bogging down the game as time accelerated, so I want to make sure that any changes are at least as performant as what we have now.

These changes seem to be significantly faster than the existing logging system, but I will admit I did somewhat limited testing. I exported my main game (almost done with hells) and imported into the game running on this branch. After letting it run for an hour it seemed to be just as quick as when the test started.

I'll try getting some benchmark data for before and after these changes while you're on vacation.

PR #177, opened 2023-02-05
Automatic code formatting

LiquidAlmond:
Most changes are from applying code formatting. I can separate the actually changes before applying the formatting rules into a separate PR if ya want.


immortalityidle (commented):
Wow that's a lot of change. I don't love the number of newlines it injects. Can you not have it line break between the tag and attributes? Other than that I don't have very strong feelings on it

LiquidAlmond:
> Wow that's a lot of change. I don't love the number of newlines it injects. Can you not have it line break between the tag and attributes? Other than that I don't have very strong feelings on it

There's a number of options.

1. We can increase the width. Currently set to 120, but can be whatever we want. We would still have the same issue if some line happens to go beyond that limit.
2. We can have Prettier ignore html files and keep the existing formatting.
3. Something I'm not thinking of...

I'm not a big fan of #1 and the html files aren't terrible, so I think I'd recommend #2.

LiquidAlmond:
Recreating without applying the formatting changes.

PR #180, opened 2023-02-11
Common state service

LiquidAlmond:
I propose making all state within the app stored within a single object and operated on exclusively by a single service like this one. By making relatively small, pure functions to be passed into the `apply` and `select` function, the code can be more easily unit tested, which will make the code more robust. This is similar to how Redux and Recoil work, but a bit more primitive.

As a side effect, all services would be reduced to a set of independent functions callable from anywhere, aside from MainLoopService, which will still need to handle ticks.

### DIFF
--- a/src/app/game-state/achievement.service.ts
+++ b/src/app/game-state/achievement.service.ts
@@ -653,7 +653,7 @@ export class AchievementService {
       description: "You have established an empire that will never fall, and a bloodline that will always inherit it.",
       hint: "Bloodline Empire.",
       check: () => {
-        return (this.homeService.home.type === HomeType.Capital && this.characterService.characterState.bloodlineRank >= 7);
+        return (this.homeService.home.type >= HomeType.Capital && this.characterService.characterState.bloodlineRank >= 7);
       },
       effect: () => {
         this.characterService.characterState.imperial = true;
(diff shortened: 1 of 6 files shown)


immortalityidle:
I'm afraid I don't understand how this makes anything better. This just seems like a lot of extra work without any significant gains. Why should the state relevant to a service not belong to that service?

LiquidAlmond:
> I'm afraid I don't understand how this makes anything better. This just seems like a lot of extra work without any significant gains. Why should the state relevant to a service not belong to that service?

immortalityidle:
Hmmm, still not sure I'm seeing the benefit. Circular dependencies seem like they're not a problem to me, is there something dangerous or performance reducing about them I'm not aware of? And I wasn't planning to add unit testing to the game ever. In any case, what you're describing doesn't seem to be what this PR actually does. This just looks like it creates an extra set of state variables hanging out in their own service.

LiquidAlmond:
Right. I don't intend for this exact PR to get merged in. I was demonstrating something that could be done to help simplify the code. Getting everything else in a state where this service would be useful would involve a lot of work and the meat of why would be lost in such a PR.

LiquidAlmond:
Canceling due to this being a much larger pain in the butt than I thought it would be with significantly less benefit.

PR #184, opened 2023-03-08
Optimize background idle

slaymaker1907:
This is kind of an odd set of changes in that it adds background music to enable playing the game without having the game focused. After about 1 minute in the background, most browsers seem to stop respecting setInterval/setTimeout regardless of the time period chosen unless audio is playing (I even tried using a WebWorker, but that did not work for Edge).

The option to play music is disabled by default and is an option in the game menu.

In addition to the audio trick, I made the scheduling a bit smarter so that we will try to run every 25ms. setInterval effectively inserts a 25ms sleep after execution which can lead to unnecessary delays.

The music is public domain from https://freepd.com/.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "immortalityidle",
-  "version": "1.1.1",
+  "version": "1.1.2",
```
immortalityidle, line 3: Please don't manually increase the version number, my build script takes care of that.

src/app/game-state/main-loop.service.ts
```diff
@@ -7,6 +7,7 @@ import { OfflineModalComponent } from '../offline-modal/offline-modal.component'
 
 const TICK_INTERVAL_MS = 25;
 const LONG_TICK_INTERVAL_MS = 500;
+const BACKGROUND_TICK_INTERVAL_MS = 1 * 1000;
```
immortalityidle, line 10: Why not just 1000?

src/app/game-state/main-loop.service.ts
```diff
@@ -20,8 +21,9 @@ export interface MainLoopProperties {
   pause: boolean;
   bankedTicks: number;
   totalTicks: number;
-  useBankedTicks: boolean;
-  scientificNotation: boolean;
+  useBankedTicks: boolean,
+  scientificNotation: boolean
```
immortalityidle, line 25: Heh, how does this not complain when you compile? punctuate this line please.

src/app/game-state/main-loop.service.ts
```diff
@@ -103,22 +107,103 @@ export class MainLoopService {
...
+    this.playMusic = properties.playMusic;
   }
 
   start() {
     if (!this.characterService) {
       this.characterService = this.injector.get(CharacterService);
     }
 
-    window.setInterval(() => {
+    // The string reason we play audio is to avoid getting deprioritized in the background.
```
immortalityidle, line 118: I'm not sure what "the string reason" means.

immortalityidle (commented):
A very interesting approach to keeping the game alive in the background. I haven't heard of this as a viable solution, but I'm up for trying anything. What's the status of the copyright of the music you included?

package.json
slaymaker1907: Ok, that's good to know. I changed it back to 1.1.1 and just reverted the package-lock.json since I assume the build script probably handles that too.

src/app/game-state/main-loop.service.ts
slaymaker1907: I was experimenting with longer times and this was an artifact of that experimentation, but 1s seems to work and longer times more often result in lag when you bring the tab up again.

src/app/game-state/main-loop.service.ts
slaymaker1907: I'm not sure why it wasn't complaining, but I've updated it. I've been testing via "npm start" since the full build takes a really long time.

src/app/game-state/main-loop.service.ts
slaymaker1907: That was a typo, I've updated the comment.

slaymaker1907:
> A very interesting approach to keeping the game alive in the background. I haven't heard of this as a viable solution, but I'm up for trying anything. What's the status of the copyright of the music you included?

@immortalityidle it is public domain from freepd.com. It isn't the best music, but to be honest I just mute the tab anyway and would recommend others do the same. Even if you put in better music, the same track is going to get tedious pretty quickly (some classical Chinese music would be more thematic, but I'll leave that up to you since it's your game).

PR #186, opened 2023-03-17
Optimize followers CPU usage

slaymaker1907:
# Follower Optimization

In particular, training followers has been optimized using Bayesian updates to only evaluate training ~365 days (or every 1/2 second) which seems to be a lot faster. If a follower only upgrades once in that period, we should only generate ~9 random numbers compared to needing to check every day before.

Additionally, most followers now only do work every 1/2 second which can yield large savings in the late game.

# Follower Power Bug in Hell

Follower power wasn't being updated on entry to hell. This should be fixed now, but I think it was a pretty widely used exploit by players so communicating about this on discord would be a good idea.

# Aptitude Optimization

Aptitude gains from BL9 are also done at the same time since we only lose a little compounding.

# Remaining Work

After these changes, the main bottlenecks seem to be combat, fields, and non-training followers actions (in descending order of impact). Combat in particular makes the other two difficult to optimize without changing game balance since it's kind of inherently serial. Maybe we could assume damage from enemies is minimal in most cases and just handle the cases where damage isn't in the slow path?

I've also noticed that CPU usage goes to about 10% without much consequence when tabbed out in conjunction with the music changes in my other PR while it remains pretty high in the foreground. I think this is probably due to CPU from Angular/DOM manipulation. Maybe this can be optimized by only updating the UI on frame subject ticks?

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -18,6 +18,7 @@ import { ChangelogPanelComponent } from './changelog-panel/changelog-panel.compo
 import { StatisticsPanelComponent } from './statistics-panel/statistics-panel.component';
 import { HellService } from './game-state/hell.service';
 import { SaveModalComponent } from './save-modal/save-modal.component';
+import { StatisticsService } from './game-state/statistics.service';
 
 @Pipe({ name: 'floor' })
 export class FloorPipe implements PipeTransform {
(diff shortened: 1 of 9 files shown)


slaymaker1907:
Also, I forgot to mention that the new statistics service should make the TPS stats persistent since Angular gets rid of the component when it is closed.

VladFurman:
Why do a binary search, instead of rolling for an arbitrary day within the range? Or even re-using the chance roll if we want to avoid calling Math.random.

slaymaker1907:
@VladFurman it's been a while since I wrote this, but the reason why it has to be binary search is because the probability distribution is not equally distributed. Followers have a higher probability of upgrading the longer they've been alive and this is supposed to account for that.

I don't think it's actually safe to reuse that chance roll either, but it might be. You'd definitely want to work it out on paper and make sure it's safe to do such that the probabilities are correct for choosing any particular day.

PR #192, opened 2023-12-16
TrainingSimplify

VladFurman:
Avoid a loop by using logarithms to calculate exact day training happens.

getProbUpgradeAfterDay = dailyFail^days

dailyFail = 1 - dailyUpgradeChance : math, static based on age, attribue, power
does not change over days

dailyFail^days = standard probability (static probability each day)

chance = dailyFail^days
log chance = days * log dailyFail
log chance / log dailyFail = days


immortalityidle:
Very cool simplification.

PR #199, opened 2025-04-16
Begin mobile layout

jimboradley:
Moved the main layout to its own component and created a new mobile layout. Set the breakpoint between the two to 500px, though that should probably be increased.

Moved the pipes into their own file, which will probably help a little with the circular imports problem.

Also added validation to loading the save file, since it was making things break for me. Check to see if the "empty" ones are sane defaults. Seemed to work fine for me but 🤷‍♂️


immortalityidle (commented):
Heh, sure, we'll give it a shot.

PR #201, opened 2025-04-26
Add activities and time. Kind of playable

jimboradley:


src/app/mobile-layout/mobile-time/mobile-time.component.html
```diff
@@ -0,0 +1,130 @@
...
+              <mat-icon tooltip="Remove this activity from the schedule." (click)="onRemoveClick(entry)" class="iconButton" aria-label="Remove Activity" aria-hidden="false">clear</mat-icon>
+            </td>
+          </tr>
+        }
+      }
+    </table>
+  </div>
+}
+</div>
+<div class="btn-row">
```
immortalityidle, line 83: Would it make more sense to have the time control buttons broken out as their own component so that they can be a header or footer for everything?

PR #204, opened 2025-05-17
Introduce carousel widgets for mobile

jimboradley:


src/app/tooltip/tooltip.directive.ts
```diff
@@ -53,9 +53,7 @@ export class TooltipDirective implements OnInit, OnDestroy {
   @HostListener('touchstart', ['$event'])
   // @ts-ignore
   show(event) {
-    event.stopPropagation();
```
jimboradley, line 56: @immortalityidle I removed this because it interferes with the ability to scroll the widgets. Is there any reason this is needed?

src/styles.less
```diff
@@ -52,6 +52,7 @@ body {
   user-select: none;
   overscroll-behavior: none;
   height: 100%;
+  width: 100%;
```
jimboradley, line 55: @immortalityidle Double check that this doesn't screw up the main panel. Possibly some sizing/scrolling capability you were relying on got busted. Same with line 97

src/app/tooltip/tooltip.directive.ts
immortalityidle: I think I put that on so that if you had multiple tooltipped elements nested in each other you should only get the popup for first one, not the whole stack.

src/styles.less
immortalityidle: I'll take a look as soon as I have time.

src/app/tooltip/tooltip.directive.ts
jimboradley: That helps a lot. Used a different method that won't supress the event for the carousel.

